PR #8, opened 2020-10-03
entgql/node: add type-resolver interface (wip)

a8m:
Don't merge yet. See https://github.com/facebook/ent/issues/811

Re-thinking about it - this won't work if users want to accepts base64-encoded ids (as suggested in GraphQL website) and the actual database-ids are ints.


giautm:
I think we need a reverse resolver from GlobalID to entity's ID and type.

a8m:
Closing, see #9

PR #12, opened 2020-10-16
Use `val` for UnmarshalGQL parameter name to avoid possible conflict with reciever name

enmand:
If the typename starts with a `v` the `receiver` template function in ent will use that as the receiver variable name for UnmarshalGQL. This PR uses `val` as the parameter for UnmarshalGQL to avoid a possible conflict.


a8m (approved):

a8m:
Thanks for your contribution @enmand

PR #14, opened 2020-10-21
Fix marshalling/unmarshalling of enums in some situations

enmand:
In the codebase we are working in, enums in ent are defined with
lower-case values, and replacing our marshal funcs with the ones
in ent-contrib for gqlgen caused errors.

This PR attempts to check the value lower-cased, if the value given
does not validate properly, and use (and validate) that value. It
also updates the marshalling to the SCREAMING_SNAKE_CASE that is
expected by GraphQL


a8m:
Hey @enmand,
How do you define the enum values in your schema? Did you see the [`NamedValues`](https://github.com/facebookincubator/ent-contrib/blob/master/entgql/internal/todo/ent/schema/todo.go#L41-L44) option?

enmand:
Hey @a8m, we're doing something like:

```
vals := []string{"val1", "val2", ...}
field.Enum("name").Values(vals...)
```

I didn't see the NamedValues thing, but I will investigate it now -- it seems like it could fix our issue without updating the template with weird string manipulation stuff. Thanks!

enmand:
Hey again @a8m, I had a look at the NamedValues -- I'm not sure it would fix the problem we're having. We were on gqlgen with custom marshalling before we started investigating ent-contrib gel integration. We defined our enum values in ent as lower-case, and our MarshalGQL upper-snake-cased things going out, and the Unmarshal lowercased things coming in from gqlgen.

To give a more concrete example, we had something like:

```
var class = []string{"fixed", "configured", "metered"}
...
    field.Enum("class").Values(class...)
```

Using `NamedValue`:

```var class = []string{
    "FIXED", "fixed",
    "CONFIGURED", "configured",
    "METERED", "metered",
}
```

just updates the identifier names, and swapping their ordering would make the stuff going into the database UPPER_CASE. Maybe the best option for us, is to override the Enum template?

a8m:
We use the `NamedValues` option to control the Go identifier and the database value, and the GraphQL value is equal to the value that is stored in the database - see how the generated enums look like in the [todo-example](https://github.com/facebookincubator/ent-contrib/blob/master/entgql/internal/todo/ent/todo/todo.go#L100-L101).

> Maybe the best option for us, is to override the Enum template?

I guess you don't want to migrate the lowercased values to uppercase, right? If that's the case, you can skip this template, use something else instead or override it with a different one.

Please let me know if you want help with that (or with the data migration).

enmand:
Thanks @a8m! For now, we're going to replace the enum template with our old template to lower-case values coming in, though I think ideally we'd migrate our values to upper-case in the database. I'll have to talk to the team here about that.

Thanks again!

a8m:
> I think ideally we'd migrate our values to upper-case in the database. I'll have to talk to the team here about that.

Sure, I can share what I did in this case:

- Added the new uppercased values with a different Go identifiers to the enum values 
- After a new version was deployed, I ran a migration command (SQL directly or using ent) to replace old values with new ones
- Removed the lowercased values from the schema and changed the Go identifiers

Thanks for your contribution and let me know if you need help with the custom template (post in facebook/ent).

PR #18, opened 2020-11-03
ID Tool

dlvhdr:
ID Tool lets you inspect ents using a web UI.


a8m:
This is currently under development (in my side), as I'm adding mutation functionality to it.

PR #23, opened 2020-12-12
Revert "entg: add Resolver annotation"

a8m:
Reverts facebookincubator/ent-contrib#22

Going to revert it until I’ll fix the multi-annotations as it requires from us to add this new annotation to all edges explicitly.

After this PR, it will generate resolvers to all edges by default, and those resolvers can be overridden on the actual resolver if necessary.


alexsn:
@giautm, I'm OK with having a `NoResolver` annonation which will allow us to explicitly disable resolver generation.

PR #26, opened 2020-12-14
feat: add NoResolver annotation

giautm:
fix typo


alexsn:
Please explain the motivation behind this change, i.e why do you need it?

giautm:
First, I want to remove code that I don't use. Second, If gqlgen check the object really has resolver implement, It will not allow to custom resolve for that field, even when function signature mismatch.

alexsn:
> Second, If gqlgen check the object really has resolver implement, It will not allow to custom resolve for that field, even when function signature mismatch.
grep for `forceResolver` in gqlgen code.

PR #34, opened 2020-12-30
examples: add todopluid which shows Prefixed ULID (PULID) example.

tmc:
This includes an example that uses prefixed ULIDs as ids. 

Prefixed ULIDs or (PULIDs) are an id encoding scheme that encodes the type of an entity being referenced in the identifier which removes the need to reference a lookup table to derive type from an ID.

This type of ID scheme may be familiar to users of Twilio APIs which include a prefix that encodes the entity type.


entgql/internal/todopulid/ent/schema/todo.go
```diff
@@ -0,0 +1,50 @@
...
+	return func() string {
+		return fmt.Sprint(pulidgql.New(prefix))
+	}
+}
+
+// Fields returns todo fields.
+func (Todo) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("id").
+			DefaultFunc(newPULIDFn("TO")),
```
a8m, line 48: Maybe use `GoType` here?  
```suggestion
		field.String("id").
			GoType(pulidgql. ID("").
			DefaultFunc(newPULIDFn("TO")),
```

a8m (commented):
Hey @tmc and thanks a lot for your contribution. 

3 comments/questions: 

1. Can you elaborate what does `pulid` mean? I didn't find any reference about it
2. What do you think about using a popular package like [oklog/ulid](https://github.com/oklog/ulid) for this example?
3. Maybe add additional info about it on some README file or create a patch for updating the doc for entgo.io

Update: 
> 1. Can you elaborate what does `pulid` mean? I didn't find any reference about it

`pulid` -> Prefixed ULID?

tmc:
> Hey @tmc and thanks a lot for your contribution.
> 
> 3 comments/questions:
> 
> 1. Can you elaborate what does `pulid` mean? I didn't find any reference about it
> 2. What do you think about using a popular package like [oklog/ulid](https://github.com/oklog/ulid) for this example?
> 3. Maybe add additional info about it on some README file or create a patch for updating the doc for entgo.io
> 
> Update:
> 
> > 1. Can you elaborate what does `pulid` mean? I didn't find any reference about it
> 
> `pulid` -> Prefixed ULID?

Yes, pulid == Prefix ULID.

oklog/ulid is a great package, I am using it in `ent/schema/pulidgql/puplidgql.go` (I'm going to rename this package to just pulid, FWIW).

I have a very small package that wraps `oklog/ulid` here: https://github.com/tmc/pulid. Let me know if you think it's more appropriate to put a small pulid package into ent-contrib instead of referencing that package.

I'd be happy to extend the docs to reference this approach, let's get the code into a better place and arrive at where a pulid package should live. Happy to follow on with a docs update.

tmc:
Added some improvements, unfortunately this has the issue of the type assertion here:

```go
	// todo.DefaultID holds the default value on creation for the id field.
	todo.DefaultID = todoDescID.Default.(func() pulid.ID)
```
failing because of the DefaultFunc in the todo.go schema requiring a func() string -- recommendation on how to address?

entgql/internal/todopulid/types.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+
+	"github.com/facebookincubator/ent-contrib/entgql/internal/todopulid/ent/schema/pulid"
+	"github.com/facebookincubator/ent-contrib/entgql/internal/todopulid/ent/todo"
+)
+
+// TypeMap maps PULID prefixes to database table names.
+var TypeMap = map[pulid.ID]string{
+	"TO": todo.Table,
+}
```
a8m, line 28: I actually expect those strings (e.g. `"TO"`) to be generated, or even generates this id2type map function. 
I can add this if you don't mind (I'll add you as a reviewer).

a8m (approved):
Looks good. Merging

entgql/internal/todopulid/types.go
tmc: Not at all! Would love the api insight -- also happy to implement given some direction.

a8m: Sure. Did you figure out why CI is falling?

a8m: > also happy to implement given some direction.

I thought about creating a `schema.Mixin` to be shared with all other schemas under the `pulid` package. For example:

```go

func MixinWithPrefix(prefix string) *Mixin {
	return &Mixin{prefix: prefix}
}

type Mixin struct {
	mixin.Schema
	prefix string
}

func (m Mixin) Fields() []ent.Field {
	return []ent.Field{
		field.String("id").
			GoType(ID("")).
			DefaultFunc(WithPrefix(m.prefix)),
	}
}

func (m Mixin) Annotations() []schema.Annotation {
	return []schema.Annotation{
		Annotation{Prefix: b.prefix},
	}
}
```

And then, we can use it as follows:

```go
// Mixin of the Todo.
func (Todo) Mixin() []ent.Mixin {
	return []ent.Mixin{
		pulid.MixinWithPrefix("TD"), // "TD" declared once.
	}
}
```

Also, define an external template that will generate the id-mapper from the mixed-in annotation:

```
// pulid/template/pulid.tmpl

var prefixMap = map[pulid.ID]string{
	{{- range $n := $.Nodes }}
		{{- if $annotation := $f.Annotations.PULID }}
			"{{ $annotation.Prefix }}": {{ $n.Package }}.Table,
		{{- end }}
	{{- end }}
}

// Exported as `ent.IDToType(...)`.
func IDToType(ctx context.Context, id pulid.ID) (string, error) {
	// ...
}
```

WDYT?

tmc: Love that direction. I’ll flesh out the idea a bit more.

a8m (approved):
Wonderful @tmc. Can you please add a few words about this example in https://entgo.io/docs/graphql/#examples?

tmc:
> Wonderful @tmc. Can you please add a few words about this example in https://entgo.io/docs/graphql/#examples?

Absolutely! I'll get a PR opened to do that.

PR #42, opened 2021-03-14
Feat/gen grpc impl skeleton

rotemtam:
This PR adds an initial version (currently useless) of a `protoc` plugin that will generate a gRPC service implementation using an ent.Client. The next PR will include actual generated method implementations for the services. This is an additional increment in implementing https://github.com/ent/ent/issues/1283


entproto/cmd/protoc-gen-entgrpc/main.go
```diff
@@ -0,0 +1,92 @@
...
+	ctx := protogen.GoImportPath("context").Ident("Context")
+	statErr := protogen.GoImportPath("google.golang.org/grpc/status").Ident("Error")
+	codeNotImp := protogen.GoImportPath("google.golang.org/grpc/codes").Ident("Unimplemented")
+	for _, me := range s.Methods {
+		g.P("// ", me.GoName, " implements ", s.GoName, "Server.", me.GoName)
+		g.P("func (svc *", s.GoName, ") ", me.GoName, "(ctx ", ctx, ", req *", me.Input.GoIdent, ") (*", me.Output.GoIdent, ", error) {")
+		g.P("return nil, ", statErr, "(", codeNotImp, ", \"error\")")
+		g.P("}")
+		g.P()
+	}
```
a8m, line 91: I guess this is some standard way to use generate files with `protogen`, but can't we use templates instead?

rotemtam: yes. this is a tool the Go team [announced](https://blog.golang.org/protobuf-apiv2) with the V2 Proto API. They use it for `protoc-gen-go` and `protoc-gen-go-grpc`. It has some very nice goodies in it (verifies the generated code, has facilities for resolving/qualifying identifiers).  You can see the code for the gRPC generator [here](https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/grpc.go)

rotemtam: Iniitially I tried to integrate templates with it, but `protogen` is a bit higher level (it modifies the code that I print to it in some ways related to imports). I can drop it and just read from stdin and write to std out, but protogen just feels very nice in the way it parses protoc options for you for example.
bottom line, instead of using templates and dancing around it I decided to embrace it instead

PR #44, opened 2021-03-15
entproto - adding FieldMap 

rotemtam:
used later in the gRPC codegen to easily describe the mapping between the ent.Schema fields and the proto desc.MessageDescriptor

### DIFF
--- a/entproto/fieldmap.go
+++ b/entproto/fieldmap.go
@@ -77,7 +77,6 @@ func (m FieldMap) Edges() []*FieldMappingDescriptor {
 type FieldMappingDescriptor struct {
 	EntField          *gen.Field
 	PbFieldDescriptor *desc.FieldDescriptor
-	ToEntExpr         string
 	IsEdgeField       bool
 	IsIDField         bool
 }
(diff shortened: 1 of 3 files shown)


entproto/fieldmap.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return nil, fmt.Errorf("entproto: could not find find %q in %q", name, entType.Name)
+}
+
+func allFields(t *gen.Type) []*gen.Field {
+	var all []*gen.Field
+	all = append(all, t.ID)
+	all = append(all, t.Fields...)
+	return all
+}
```
a8m, line 132:
```suggestion
func extractEntFieldByName(entType *gen.Type, name string) (*gen.Field, error) {
	if name == entType.ID.Name {
		return entType.ID, nil
	}
	for _, fld := range allFields(entType) {
		if fld.Name == name {
			return fld, nil
		}
	}
	return nil, fmt.Errorf("entproto: could not find find %q in %q", name, entType.Name)
}
```

Easier to follow this way IMO. Also, this reduces unnecessary allocation in `allFields`.

entproto/fieldmap.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}
+	return out
+}
+
+// FieldMappingDescriptor describes the mapping from a protobuf field descriptor to an ent Schema field
+type FieldMappingDescriptor struct {
+	EntField          *gen.Field
+	PbFieldDescriptor *desc.FieldDescriptor
+	ToEntExpr         string
```
a8m, line 80: Later usage?

entproto/internal/entprototest/fieldmapper_test.go
```diff
@@ -0,0 +1,37 @@
...
+	suite.Assert().NotNil(userName.PbFieldDescriptor)
+	suite.Assert().NoError(err)
+	suite.Assert().EqualValues("UserName", userName.PbStructField())
+
+	id, ok := mp["id"]
+	suite.Require().True(ok)
+	suite.Assert().True(id.IsIDField)
+	suite.Assert().False(id.IsEdgeField)
+	suite.Assert().EqualValues("Id", id.PbStructField())
+}
```
a8m, line 37: Add test with `IsEdgeField` as well?

a8m (approved):

entproto/fieldmap.go
rotemtam: nope. removing

PR #45, opened 2021-03-15
Feat/entproto printtemplate helper

rotemtam:
Notice the PR here is to feat/entproto-fieldmap, this is just for convenience once that is merged i will change it to master

### DIFF
--- a/entproto/cmd/protoc-gen-entgrpc/template.go
+++ b/entproto/cmd/protoc-gen-entgrpc/template.go
@@ -1,3 +1,17 @@
+// Copyright 2019-present Facebook
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package main
 
 import "fmt"
--- a/entproto/cmd/protoc-gen-entgrpc/template.go
+++ b/entproto/cmd/protoc-gen-entgrpc/template.go
@@ -14,52 +14,44 @@
 
 package main
 
-import "fmt"
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/hashicorp/go-multierror"
+	"google.golang.org/protobuf/compiler/protogen"
+)
 
 // printTemplate is a utility function to make working with protogen have a more declarative interface.
-// It receives a protogenPrinter (in practice a *protogen.GeneratedFile), a template string with placeholder
-// formatted like "%(variableName)" and a tmplValues map containing the values that should be replaced when
-// the template is rendered.
+// It receives a *protogen.GeneratedFile, a template string with placeholder formatted like "%(variableName)"
+// and a tmplValues map containing the values that should be replaced when the template is rendered.
 //
 // Instead of:
-//	g.P("func ", svcName, "(", paramName, " string)")
+//	g.P("func New", svcName, "(p string) *", protogen.GoImportPath("context").Ident("Context"))
 // We can use
-//	printTemplate(g, "func %(svcName)(%(paramName) string)
-func printTemplate(printer protogenPrinter, template string, values tmplValues) error {
-	var inToken bool
-	var buf string
-	var output []interface{}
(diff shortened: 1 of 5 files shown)


entproto/cmd/protoc-gen-entgrpc/template.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+		}
+	}
+	if inToken {
+		return fmt.Errorf("entproto: corrupt template, must close parenthesis")
+	}
+	output = append(output, buf)
+	printer.P(output...)
+	return nil
+}
```
a8m, line 64: Nice, you can achieve something similar with regexp:

```go
re := regexp.MustCompile(`(%\(.+?\))`)
out := re.ReplaceAllStringFunc(tmpl, func(s string) string {
	switch v := values[s]; v.(type) {
	case protogen.GoIdent:
		return g.QualifiedGoIdent(v) // The P(..) magic for Go identifiers. 
	default:
		return fmt.Sprint(v)
	}
}))
g.P(out)
```

entproto/cmd/protoc-gen-entgrpc/template.go
```diff
@@ -0,0 +1,78 @@
...
+// Instead of:
+//	g.P("func ", svcName, "(", paramName, " string)")
+// We can use
+//	printTemplate(g, "func %(svcName)(%(paramName) string)
+func printTemplate(printer protogenPrinter, template string, values tmplValues) error {
+	var inToken bool
+	var buf string
+	var output []interface{}
+	for _, c := range template {
+		str := string(c)
```
a8m, line 33: Why this conversion is needed? You can compare runes as well (e.g. `c == '('`)

entproto/cmd/protoc-gen-entgrpc/template.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+		} else {
+			if str == "%" {
+				inToken = true
+				output = append(output, buf)
+				buf = str
+			} else {
+				buf += str
+			}
+		}
```
a8m, line 56:
```suggestion
		} else if str == "%" {
			inToken = true
			output = append(output, buf)
			buf = str
		} else {
			buf += str
		}
```

Or `switch` with `default`.

entproto/cmd/protoc-gen-entgrpc/template.go
```diff
@@ -0,0 +1,78 @@
...
+// printTemplate is a utility function to make working with protogen have a more declarative interface.
+// It receives a protogenPrinter (in practice a *protogen.GeneratedFile), a template string with placeholder
+// formatted like "%(variableName)" and a tmplValues map containing the values that should be replaced when
+// the template is rendered.
+//
+// Instead of:
+//	g.P("func ", svcName, "(", paramName, " string)")
+// We can use
+//	printTemplate(g, "func %(svcName)(%(paramName) string)
+func printTemplate(printer protogenPrinter, template string, values tmplValues) error {
```
a8m, line 28: This interface is not really needed. Instead, pass a print function instead. 

```go
func printTemplate(print func(...interface{}), ....) error

// Or:
type PrintFunc func(...interface{})

func printTemplate(print PrintFunc, ....) error 
```

entproto/cmd/protoc-gen-entgrpc/template.go
rotemtam: this is a really great suggestion. i ended up doing just that. my prime thought is that its much easier to change if we need to add more subtleties to the "template language"

a8m: Great 😃

a8m (approved):

entproto/cmd/protoc-gen-entgrpc/template.go
```diff
@@ -0,0 +1,66 @@
...
+//
+// Instead of:
+//	g.P("func New", svcName, "(p string) *", protogen.GoImportPath("context").Ident("Context"))
+// We can use
+//	printTemplate(g, "func New%(svcName)(p string) *%(ctx)", tmplValues{
+//		"svcName": "UserService",
+//		"ctx": protogen.GoImportPath("context").Ident("Context"),
+//	})
+func printTemplate(g *protogen.GeneratedFile, tmpl string, values tmplValues) error {
+	re := regexp.MustCompile(`(%\(.+?\))`)
```
a8m, line 37: Just one comment - unlike Python, the `regexp` package doesn't cache these regular-expressions.
Move this variable to the top-level declarations:

```go
var re = regexp.MustCompile(`(%\(.+?\))`)

func printTemplate(g *protogen.GeneratedFile, tmpl string, values tmplValues) error {
	// ...
}
```

PR #47, opened 2021-03-16
implement the mapper func gen code ent -> proto 

rotemtam:
generate a toProto func for the service entity type


entproto/cmd/protoc-gen-entgrpc/cast.go
```diff
@@ -0,0 +1,53 @@
...
+		ident := g.pbEnumIdent(fld)
+		methodName := "toProto" + ident.GoName
+		return methodName, nil
+	case dpb.FieldDescriptorProto_TYPE_MESSAGE:
+		if pbd.GetMessageType().GetFullyQualifiedName() == "google.protobuf.Timestamp" {
+			newTS := protogen.GoImportPath("google.golang.org/protobuf/types/known/timestamppb").Ident("New")
+			return newTS, nil
+		} else {
+			return nil, fmt.Errorf("entproto: no mapping for pb message type %q", pbd.GetFullyQualifiedName())
+		}
```
a8m, line 49: This flow is preferred is more idiomatic: 
```suggestion
		If name := pbd.GetMessageType(); name != "google.protobuf.Timestamp" {
			return nil, fmt.Errorf("entproto: no mapping for pb message type %q", name)
		}
		return protogen.GoImportPath("google.golang.org/protobuf/types/known/timestamppb").Ident("New")
```

entproto/cmd/protoc-gen-entgrpc/main.go
```diff
@@ -202,11 +227,16 @@ func (g *serviceGenerator) generateMethod(me *protogen.Method) error {
...
+	typeName := strings.TrimSuffix(s.GoName, "Service")
 	for _, gt := range g.Nodes {
 		if gt.Name == typeName {
 			return typeName, nil
 		}
 	}
 	return "", fmt.Errorf("entproto: type %q of service %q not found in graph", typeName, s.GoName)
 }
+
+func (g *serviceGenerator) entSymbol(subpath string, ident string) protogen.GoIdent {
```
a8m, line 239: entSymbol -> entIdent (it returns a Go identifier)

entproto/fieldmap.go
```diff
@@ -79,6 +79,20 @@ func (m FieldMap) Edges() []*FieldMappingDescriptor {
...
+func (m FieldMap) Enums() []*FieldMappingDescriptor {
+	var out []*FieldMappingDescriptor
+	for _, f := range m {
+		if f.IsEnumFIeld {
+			out = append(out, f)
+		}
+	}
+	sort.Slice(out, func(i, j int) bool {
+		return out[i].PbStructField() < out[j].PbStructField()
+	})
```
a8m, line 95: I saw this sorting a few times here - why it's needed? Why not keep the same order as in the schema?

entproto/internal/todo/ent/proto/entpb/user_service_test.go
```diff
@@ -0,0 +1,47 @@
...
+	ts := time.Now()
+	created := client.User.Create().
+		SetUserName("rotemtam").
+		SetExp(100).
+		SetPoints(1000).
+		SetStatus("active").
+		SetJoined(ts).
+		SaveX(context.Background())
+
+	pbUser := toProtoUser(created)
```
a8m, line 40: Are we testing a private function here? Is it temporary?

rotemtam: yes, we can kick it out in the next PR since it is implicitly tested by the public service implementation

entproto/fieldmap.go
rotemtam: b/c:
```go
// FieldMap contains a mapping between the field's name in the ent schema and a FieldMappingDescriptor.
type FieldMap map[string]*FieldMappingDescriptor`
```
its a map, which doesn't have stable order when ranged

a8m (approved):

PR #48, opened 2021-03-17
Generate the Create gRPC method

rotemtam:


entproto/cmd/protoc-gen-entgrpc/cast.go
```diff
@@ -51,3 +52,27 @@ func (g *serviceGenerator) castToProtoFunc(fld *entproto.FieldMappingDescriptor)
...
 	}
 }
+
+func (g *serviceGenerator) castToEntFunc(fd *entproto.FieldMappingDescriptor) (interface{}, error) {
+	var fld *gen.Field
+	if fd.IsEdgeField {
+		fld = fd.EntEdge.Type.ID
+	} else {
+		fld = fd.EntField
+	}
```
a8m, line 62: Not mandatory, but I usaully do:

```suggestion
	fld := fd.EntField
	if fd.IsEdgeField {
		fld = fd.EntEdge.Type.ID
	}
```

entproto/cmd/protoc-gen-entgrpc/methods.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"entgo.io/ent/entc/gen"
+	"google.golang.org/protobuf/compiler/protogen"
+)
+
+var (
+	camel = gen.Funcs["camel"].(func(string) string)
+)
+
+func (g *serviceGenerator) generateCreateMethod() error {
```
a8m, line 26: Yay!!! 😍

entproto/internal/entprototest/fieldmapper_test.go
```diff
@@ -41,6 +41,8 @@ func (suite *AdapterTestSuite) TestFieldMap() {
 	assert.EqualValues("BlogPosts", blogPosts.PbStructField())
 	assert.False(blogPosts.IsIDField)
 	assert.True(blogPosts.IsEdgeField)
+	assert.EqualValues("Id", blogPosts.EdgeIDPbStructField())
```
a8m, line 44: Is it common to write the identifier field this way, and not `ID`?

a8m (approved):
Wow @rotemtam! That's really great.


Thanks a lot for pushing this

entproto/internal/entprototest/fieldmapper_test.go
rotemtam: Yes this ist he way protoc generated Go code from a field named `id`

PR #53, opened 2021-03-31
entgql/template: add support for id ordering

alexsn:
Calling `$node.Fields` doesn't include ID field so manually append it when building `$orderFields` list.
Fixes https://github.com/ent/ent/issues/1359


a8m (approved):
👌🏻

PR #54, opened 2021-04-01
Feat/entproto support UUID

rotemtam:
First PR in a sequence of changes to support UUID type fields in generated gRPC servers in order to address https://github.com/ent/ent/issues/1402


a8m (approved):
👍

PR #58, opened 2021-04-11
Feat/proto custom type

rotemtam:
In prep for https://github.com/ent/ent/discussions/1350, adding some flexibilty for users to modify the type mapping between ent and pb.


entproto/field.go
```diff
@@ -35,13 +36,29 @@ func Field(num int, options ...FieldOption) schema.Annotation {
...
+		p.OverrideType = typ
+	}
+}
+
+// OverrideTypeName sets the pb descriptors type name, needed if the OverrideType attribute is TYPE_ENUM or TYPE_MESSAGE.
+func OverrideTypeName(n string) FieldOption {
+	return func(p *pbfield) {
+		p.OverrideTypeName = n
+	}
+}
```
a8m, line 60: Looks good, but remove the `Override` string from this function. 

Also, consider adding a usage example in the godoc. I use it a lot. 

```suggestion
func Type(typ descriptorpb.FieldDescriptorProto_Type) FieldOption {
	return func(p *pbfield) {
		p.Type = typ
	}
}

// TypeName sets the pb descriptors type name, needed if the OverrideType attribute is TYPE_ENUM or TYPE_MESSAGE.
func TypeName(n string) FieldOption {
	return func(p *pbfield) {
		p.TypeName = n
	}
}
```

a8m (approved):

PR #59, opened 2021-04-14
entproto/adapter: support optional and nillable fields by representin…

rotemtam:
…g them as pb wrapper types

related to https://github.com/ent/ent/discussions/1350

gRPC  codegen still does not support this, ill add this in another PR


entproto/internal/entprototest/ent/schema/optionals.go
```diff
@@ -0,0 +1,62 @@
...
+		field.Bytes("bytes_field").
+			Optional().
+			Annotations(entproto.Field(7)),
+		field.UUID("uuid_field", uuid.New()).
+			Optional().
+			Annotations(entproto.Field(8)),
+		field.Time("time_field").
+			Optional().
+			Annotations(entproto.Field(9)),
+	}
```
a8m, line 55: Why the `_field` suffix? Maybe prefix them with `optional_` if you prefer not to use the type name as the field name.

a8m (approved):

PR #60, opened 2021-04-18
entproto/cmd/protoc-gen-entgrpc: support rendering wrapped calls

rotemtam:
needed for supporting two step castings such as wrapperspb.Int32(int32(1))

### DIFF
--- a/entproto/cmd/protoc-gen-entgrpc/template.go
+++ b/entproto/cmd/protoc-gen-entgrpc/template.go
@@ -17,6 +17,7 @@ package main
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/hashicorp/go-multierror"
 	"google.golang.org/protobuf/compiler/protogen"
@@ -48,6 +49,8 @@ func printTemplate(g *protogen.GeneratedFile, tmpl string, values tmplValues) er
 		switch p := val.(type) {
 		case protogen.GoIdent:
 			return g.QualifiedGoIdent(p) // The P(..) magic for Go identifiers.
+		case wrappedCalls:
+			return renderWrappedCalls(g, p)
 		default:
 			return fmt.Sprint(p)
 		}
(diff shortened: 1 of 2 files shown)


entproto/cmd/protoc-gen-entgrpc/template.go
```diff
@@ -68,3 +71,23 @@ func (t tmplValues) retrieve(token string) (interface{}, error) {
 	}
 	return k, nil
 }
+
+type wrappedCalls struct {
```
a8m, line 75: The name `calls` is confusing because conversion is not a function call. Do you find it clearer than this?

```go
g.Tmpl("v := %(a)(%(b))", tmplValues{
	"a": protogen.GoImportPath("golang.google.com/protobuf/wrapperspb").Ident("Int32"),
	"b": "int32(x)",
})
```

rotemtam: yes because we need to support 3 cases in this template:
1. ent field type == pb field type - no conversion/function
2. simple conversion - ( int32(1) )
3. two step conversion wrapperspb.Int32(int32(1))
having this struct that can represent all cases makes the template and passing the instructions around the code cleaner

maybe if we call it `typeConversion`, and the member field is called `ops` or something like that?

a8m: > 1. ent field type == pb field type - no conversion/function

Why do you need it in this case, and how do you use it in this example.

> 2. simple conversion - ( int32(1) )
> 3. two step conversion wrapperspb.Int32(int32(1))
>    having this struct that can represent all cases makes the template and passing the instructions around the code cleaner
> 
> maybe if we call it `typeConversion`, and the member field is called `ops` or something like that?

Not sure. The first one is a type conversion and the second one is a function call. Maybe let's just make it more explicit for the use-case that you use in the template (do you have any examples to share? maybe in another PR)? Also, `arguments` should be a single `argument`, right?

rotemtam:
Rethinking this, this logic doesn't fit well in the templating layer and should be part of a sort of a "codec" layer (which has an initial existence in cast.go) which should more holistically do the conversions between proto/ent. Esp, if we will later want to extract it to a common layer which can be used by GraphQL schema gen, etc. 

I'm closing this and will send a sketch tomorrow 

cc @a8m

PR #70, opened 2021-05-03
schemast: support schema.Annotation in fields, edges and mutators

rotemtam:


schemast/annotation.go
```diff
@@ -56,6 +56,62 @@ func Annotation(annot schema.Annotation) (ast.Expr, bool, error) {
...
+		if r.Name == "nil" {
+			stmt.Results = []ast.Expr{
+				newAnnotSliceWith(newAnnot),
+			}
+			return nil
+		}
+		return fmt.Errorf("schemast: unexpected ident. expected nil got %s", r.Name)
+	case *ast.CompositeLit:
+		r.Elts = append(r.Elts, newAnnot)
+		return nil
```
a8m, line 83: This code is repeated itself in other `Append.*` methods. Consider reusing this block in all other methods (can be done in an additional PR).

schemast/annotation.go
```diff
@@ -56,6 +56,62 @@ func Annotation(annot schema.Annotation) (ast.Expr, bool, error) {
...
+	return &ast.CompositeLit{
+		Type: &ast.ArrayType{
+			Elt: &ast.SelectorExpr{
+				X:   ast.NewIdent("schema"),
+				Sel: ast.NewIdent("Annotation"),
+			},
+		},
+		Elts: annotations,
+	}
+}
```
a8m, line 113: Ditto. The only difference is the identifiers.

schemast/field.go
```diff
@@ -161,12 +161,25 @@ func fromSimpleType(desc *field.Descriptor) (*ast.CallExpr, error) {
...
+			}
+			if !shouldAdd {
+				continue
+			}
+			annots = append(annots, a)
+		}
+		if len(annots) > 0 {
+			builder.method("Annotations", annots...)
+		}
+	}
```
a8m, line 179: ditto

schemast/printer_test.go
```diff
@@ -139,3 +139,9 @@ func (p *printTest) getType(name string) *gen.Type {
 	}
 	return nil
 }
+
+func (p *printTest) getContents(fname string) string {
```
a8m, line 143:
```suggestion
func (p *printTest) contents(fname string) string {
```

The `get` prefix is not needed here. It's fine for the `getType` above because `type` is a keyword.

a8m (approved):

schemast/annotation.go
rotemtam: 👍

PR #73, opened 2021-05-04
schemast: add support for appending indices in mutations

rotemtam:


schemast/index.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"go/ast"
+
+	"entgo.io/ent"
+	"entgo.io/ent/schema/index"
+)
+
+// Index converts a *index.Descriptor back into an *ast.CallExpr of the ent Index package that can be used
+// to construct it.
+func Index(desc *index.Descriptor) (*ast.CallExpr, error) {
```
a8m, line 26:
```suggestion
// Index converts an *index.Descriptor back into an *ast.CallExpr
// of the ent Index package that can be used to construct it.
func Index(desc *index.Descriptor) (*ast.CallExpr, error) {
```

schemast/index.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if desc.StorageKey != "" {
+		idx.method("StorageKey", strLit(desc.StorageKey))
+	}
+	if len(desc.Edges) > 0 {
+		var edges []ast.Expr
+		for _, edg := range desc.Edges {
+			edges = append(edges, strLit(edg))
+		}
+		idx.method("Edges", edges...)
```
a8m, line 39:
```suggestion
		var edges []ast.Expr
		for _, e := range desc.Edges {
			edges = append(edges, strLit(e))
		}
		idx.method("Edges", edges...)
```

You can use `make` btw.

a8m (approved):

PR #77, opened 2021-05-14
Correctly filter fields that are not annotation with OrderField

cliedeman:


entgql/template/pagination.tmpl
```diff
@@ -217,7 +217,9 @@ const (
 		{{- if not $f.Type.Comparable }}
 			{{ fail (printf "annotated field %s.%s must be comparable" $node.Name $f.Name) }}
 		{{- end }}
+		{{- if $annotation.OrderField }}
 		{{ $orderFields = append $orderFields $f }}
+		{{- end }}
```
a8m, line 222:
```suggestion
		{{- if $annotation.OrderField }}
			{{ $orderFields = append $orderFields $f }}
		{{- end }}
```

a8m (approved):
Looks good. Thanks for fixing this @cliedeman.

Please add the indentation and I'll merge it.

PR #78, opened 2021-05-15
Gqlgen plugin

cliedeman:
Improve ent + gqlgen interop

Features

- [x] Automatically map objects
- [x] Generate Relay objects
- [x] Custom scalars
- [x] Map Pagination (Order by) Input Objects
- [ ] Map Enums
- [ ] Map Edges
- [ ] Rename fields
- [ ] ~Support Default Values~  Only makes sense for inputs


cliedeman:
@a8m do you have some thoughts on the annotations?

```go
type Annotation struct {
	// OrderField is the ordering field as defined in graphql schema.
	OrderField string `json:"OrderField,omitempty"`
	// Bind implies the edge field name in graphql schema
	// is equivalent to the name used in ent schema.
	Bind bool `json:"Bind,omitempty"`
	// Mapping is the edge field names as defined in graphql schema.
	Mapping []string `json:"Mapping,omitempty"`
	// Skip exclude the type
	Skip bool `json:"Skip,omitempty"`
	// RelayConnection expose this node as a relay connection
	RelayConnection bool `json:"RelayConnection,omitempty"`
	// GqlName provide alternative name. see: https://gqlgen.com/config/#inline-config-with-directives
	GqlName string `json:"GqlName,omitempty"`
	// GqlType override type
	GqlType string `json:"GqlType,omitempty"`
	// GqlImplements extra interfaces that are implemented
	GqlImplements []string `json:"GqlImplements,omitempty"`
	// GqlScalarMappings defines custom scalars mappings, scalars will also be created automatically
	GqlScalarMappings map[string]string `json:"GqlScalarMappings,omitempty"`
}
```

It's already got quite a few fields and I forsee having to add more

- ForceResolver
- DataLoader (future addition)

Should we consider creating a second annotation or nesting some of the related annotations under another object?

delaneyj:
All this sounds amaazing.  What's the PR's current status?

baranyildirim:
Also curious about this. I'm trying to go from a GraphQL schema to Ent structs, but I don't think it is possible right now.

StevenACoffman:
@cliedeman hi! I maintain gqlgen, and we've updated a lot for `v0.15.1`, but I'd be very interested in hearing if there's anything else we can do to support your work here.

giautm:
After this commit https://github.com/ent/contrib/commit/a551ba9984843cc1201f5dc2090185ad5ca73b78. `entgql` now can load GQLGen's plugins when running.

I think we should have only 1 process to generate code, because it may cause the conflict if `gqlgen` running before/after `entc`

StevenACoffman:
So with Ent v0.10.0 out, I wonder if @a8m might have a chance to weigh in more on this PR. 

I also noticed, that the plan for Ent v0.11.0 includes:
> Reimplementing the GraphQL integration to be fully compatible with the Relay spec. Supporting generating GraphQL assets (schemas or full servers) from Ent schemas.

Is there anything gqlgen might change or implement to better support Ent?

cliedeman:
@StevenACoffman thanks for reaching out. Really nice to see those gqlgen releases rolling out.

I only recall two issues when working on this plugin.

1. The plugins, EarlySourceInjector for example cannot return an error. I had to move some stuff around and do all the initialization logic upfront but I think some panics still crept into the code beacause I couldn't return an error
2. `@goField(forceResolver: Boolean)` This is handy in quite a few places when combining ent + gqlgen but it overrides the model and creates a copy. I would like in some cases to create a custom resolver but still return the ent type. 

Do you think I should create issues for these?

It looks like @giautm has taken over this PR in #78 so I will close this

Ciaran

StevenACoffman:
@cliedeman Yes, please create issues for those both! The second one in particular I'm not sure how to best accommodate that. Ideas (and PRs) are always welcome if anyone wants to take a stab at it!

cliedeman:
@StevenACoffman I logged the following tickets. 1 and 2 should be relatively easy. The force resolver one is a litlte bit harder but shouldn't be too bad, its mostly about deciding on a good api first
https://github.com/99designs/gqlgen/issues/1990
https://github.com/99designs/gqlgen/issues/1991
https://github.com/99designs/gqlgen/issues/1992

cliedeman:
@StevenACoffman those pull requests are pretty big breaking changes. I usually expect some sort of approval or buy in from maintainers before starting work on them. Or should we just create the Pull requests and start the discussion there?

StevenACoffman:
Well, you had mentioned that the first two would be relatively easy, so I would think a proof-of-concept Pull Request would then allow everyone to explore accomplishing the same thing but **without** breaking backward compatibility. For instance, having the new behavior behind some sort of config switch or alternate API.

PR #79, opened 2021-05-17
Correct OrderField template validation

cliedeman:
The Previous fix in #77 was incomplete.

It only fixed a field that is not comparable without an annotation
```
field.Bytes("blob")
```

but not an incomparable field with Annotation
```
field.Bytes("blob").Annotations(entgql.Annotation{})
```


a8m (approved):
Thanks @cliedeman

PR #80, opened 2021-05-20
Correct OrderField template validation

cliedeman:
Not sure what happened but the previous fix is not on master


a8m:
Thanks @cliedeman 😃

PR #84, opened 2021-06-12
entgql: revert annotation json tags from #76 

a8m:
Existing (external) templates stopped working after #76, and this change should fix this.


entgql/annotation.go
```diff
@@ -22,14 +22,14 @@ import (
...
 	// is equivalent to the name used in ent schema.
-	Bind bool `json:"bind,omitempty"`
+	Bind bool `json:"Bind,omitempty"`
 	// Mapping is the edge field names as defined in graphql schema.
-	Mapping []string `json:"mapping,omitempty"`
+	Mapping []string `json:"Mapping,omitempty"`
 	// Skip exclude the type
-	Skip bool `json:"skip,omitempty"`
+	Skip bool `json:"Skip,omitempty"`
 }
```
a8m, line 33: cc @cliedeman

cliedeman: Thanks @a8m 
Should I make new fields snake_case or keep them camel case to match existing fields

PR #85, opened 2021-06-19
entproto/cmd/protoc-gen-entgrpc: do not set zero value to nillable fi…

rotemtam:
…elds on Create/Update if nil is supplied.

Generated Create/Update methods will now do something like:
```go
if user.GetOptBool() != nil {
		m.SetOptBool(user.GetOptBool().GetValue())
}
```
Since the optional fields are already wrapper types (struct that can are nil if not supplied) this works neatly on top of that.


a8m (approved):

a8m:
Thanks for fixing this @rotemtam.

PR #89, opened 2021-06-29
feat: value scanner type support

tarrencev:
this is a proof of concept of how we can integrate valuer type conversion. we could do something similar on the scanner side. i wanted to open this sooner rather than later for feedback. if you agree with the approach, i can continue work on it to support all value scanner types.

one suggestion i would make is to remove the runtime panics and instead return an appropriate grpc error status when conversion fails (i.e. `InvalidArgument` on requests/`FailedPrecondition` on responses).

### DIFF
--- a/entproto/cmd/protoc-gen-entgrpc/converter.go
+++ b/entproto/cmd/protoc-gen-entgrpc/converter.go
@@ -37,8 +37,14 @@ func (g *serviceGenerator) newConverter(fld *entproto.FieldMappingDescriptor) (*
 	out := &converter{}
 	pbd := fld.PbFieldDescriptor
 	switch pbd.GetType() {
+	case dpb.FieldDescriptorProto_TYPE_BYTES:
+		if fld.EntField.Type.Valuer() {
+			out.toProtoConstructor = protogen.GoImportPath("entgo.io/contrib/entproto/runtime").Ident("MustValueToBytes")
+		} else if err := basicTypeConversion(fld.PbFieldDescriptor, fld.EntField, out); err != nil {
+			return nil, err
+		}
 	case dpb.FieldDescriptorProto_TYPE_BOOL, dpb.FieldDescriptorProto_TYPE_STRING,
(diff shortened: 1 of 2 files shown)


rotemtam:
HI @tarrencev !

Thanks a lot for pushing this. I'm very happy to collaborate with you on this. This is a good start, here are some more things to consider:

### The dev loop

The tests for `protoc-gen-entgrpc` are in `entproto/internal/todo`. the dev flow should be:
* Re-build and install the protoc plugin
* Regenerate the ent project (which invokes entproto to generate proto files and regenerates the service definition and implementation) 
* Run the unit tests (these are unit tests against the generated code)

When working on `protoc-gen-entgrpc`, I usually run this command:
```console
go get entgo.io/contrib/entproto/cmd/protoc-gen-entgrpc && go generate ./internal/todo/... && go test ./internal/todo/ent/proto/...
```

As part of this PR we should add to one of the schemas entproto/internal/todo (for example, https://github.com/ent/contrib/blob/master/entproto/internal/todo/ent/schema/attachment.go) a field which would represent the use case - for example a big.Int field represented by `[]byte` in pb. The respective unit test should be updated to test the full flow of reading and writing from the field.


### Input Validation

> one suggestion i would make is to remove the runtime panics and instead return an appropriate grpc error status when conversion fails (i.e. InvalidArgument on requests/FailedPrecondition on responses).

You are absolutely correct. If you take a closer look at the code you will see that in cases where such panics can occur, a validator is generated and run before the actual conversion, for example:
https://github.com/ent/contrib/blob/668b53bb06ab83b495e780525dc155b64538e896/entproto/internal/todo/ent/proto/entpb/entpb_user_service.go#L81-L98

This bit of logic is handled in https://github.com/ent/contrib/blob/master/entproto/cmd/protoc-gen-entgrpc/validation.go.

This logic should be modified to generate a call to a validation function.

### Proto to Ent

If you examine the code for defining "converters" for specific fields you will see that they go in both directions, your PR currently begins to touch on the "to proto" side, but we need to go the other way around and define to "to ent" conversion. 

### Collaboration

I hope this gives enough context and expectations on the feature. Let's continue the discussion here on in the slack channel 🙏

PR #96, opened 2021-07-04
entproto: valuescanner support

tarrencev:
Opening this early for feedback on uuid changes necessary to integrate it using `ValueScanner`. We switch to using `string` to represent UUID rather than bytes, allowing us to generalize the implementation.


tarrencev:
Something I am struggling with:

When storing ValueScanner types in the proto, we need to someone decide what type to use. Right now, the type is derived based on the Field type. But if the user specifies a custom `GoType` then it is stored using the `Valuer` type (`driver.Value`). How do we figure out the `driver.Value` type to set the proto type when generating the protos?

rotemtam:
Hey @tarrencev ,

Re providing the proto type annotation, this is possible today:
https://github.com/ent/contrib/blob/e16031b69b23333c597e278f7472040b7ac990eb/entproto/field.go#L48-L60

entproto/cmd/protoc-gen-entgrpc/converter.go
```diff
@@ -65,7 +66,21 @@ func (g *serviceGenerator) newConverter(fld *entproto.FieldMappingDescriptor) (*
 	if fld.IsEdgeField {
 		efld = fld.EntEdge.Type.ID
 	}
+
 	switch {
+	case efld.IsUUID() && efld.Type.ValueScanner():
+		out.toEntScannerConstructor = protogen.GoImportPath("github.com/google/uuid").Ident("UUID")
+	case efld.HasGoType() && efld.Type.ValueScanner():
+		// Ident returned from ent already has the packagename prefixed. Strip it since `g.QualifiedGoIdent`
+		// adds it back.
```
tarrencev, line 75: @rotemtam @a8m please let me know if there is a cleaner way to do this

entproto/cmd/protoc-gen-entgrpc/converter.go
```diff
@@ -136,13 +163,32 @@ func (g *serviceGenerator) renderToProto(conv *converter, varName, ident string)
...
 		}
-		`, varName, g.QualifiedGoIdent(conv.toProtoTryConstructor), ident)
+
+		%sTyped, ok := %sValue.(%s)
+		if !ok {
+			return nil, %s("casting value to %s")
+		}
+
+		%s := %s(%sTyped)
+		`, varName, ident, varName, varName, conv.toProtoValuer, "%(newError)", conv.toProtoValuer, varName, g.QualifiedGoIdent(conv.toProtoConstructor), varName)
```
tarrencev, line 179: these are getting pretty complex. i opted here to duplicate the logic rather than increase the conditional logics complexity (since it is hard to reason about).

i think as a followup we should improve the templating system to allow us to use them here. that way we can have a single way of templating rather than combining fmt and the imperative templating helper.

entproto/cmd/protoc-gen-entgrpc/converter.go
rotemtam: I think that `protogen.GoImportPath(split[1]).Ident(efld.Type.PkgPath)` should be ok?

entproto/cmd/protoc-gen-entgrpc/converter.go
```diff
@@ -65,7 +66,21 @@ func (g *serviceGenerator) newConverter(fld *entproto.FieldMappingDescriptor) (*
 	if fld.IsEdgeField {
 		efld = fld.EntEdge.Type.ID
 	}
+
 	switch {
+	case efld.IsUUID() && efld.Type.ValueScanner():
+		out.toEntScannerConstructor = protogen.GoImportPath("github.com/google/uuid").Ident("UUID")
```
a8m, line 72: Please use the `gen.Field.Type.PkgPath` instead.

entproto/cmd/protoc-gen-entgrpc/converter.go
```diff
@@ -65,7 +66,21 @@ func (g *serviceGenerator) newConverter(fld *entproto.FieldMappingDescriptor) (*
...
+		// Ident returned from ent already has the packagename prefixed. Strip it since `g.QualifiedGoIdent`
+		// adds it back.
+		split := strings.Split(efld.Type.Ident, ".")
+		out.toEntScannerConstructor = protogen.GoIdent{GoName: split[1], GoImportPath: protogen.GoImportPath(efld.Type.PkgPath)}
+	case efld.IsBool() && efld.Type.ValueScanner():
+		out.toEntScannerConversion = "bool"
+	case efld.IsBytes() && efld.Type.ValueScanner():
+		out.toEntScannerConversion = "[]byte"
+	case efld.IsString() && efld.Type.ValueScanner():
+		out.toEntScannerConversion = "string"
```
a8m, line 83: Please replace all the `efld.Type.ValueScanner()` checks with one check. 
```suggestion
	switch {
	case !efld.Type.ValueScanner():
		switch {
		case efld.IsBool(), efld.IsBytes(), efld.IsString():
		case efld.Type.Numeric():
			// ..
		default:
			// ..
		}
	case efld.IsUUID():
		// ...
	case efld.HasGoType():
		// ...
	case efld.IsBool():
		out.toEntScannerConversion = "bool"
	case efld.IsBytes():
		out.toEntScannerConversion = "[]byte"
	case efld.IsString():
		out.toEntScannerConversion = "string"
	}
```

entproto/cmd/protoc-gen-entgrpc/converter.go
```diff
@@ -194,3 +271,30 @@ var wrapperPrimitives = map[string]string{
 	"google.protobuf.StringValue": "string",
 	"google.protobuf.BytesValue":  "[]byte",
 }
+
+func implements(r *field.RType, typ reflect.Type) bool {
```
tarrencev, line 275: This is taken from ent/ent. I ported it over since I don't think it makes sense to make a public api in `ent` and might be useful for other checks here i.e. `MarshallProto` interface.

PR #97, opened 2021-07-05
entgql: change where input to camel-case

giautm:


giautm:
Hừm? Why the check run failure?

PR #99, opened 2021-07-07
Added missing type

cliedeman:
Need to add very secret to other examples otherwise they break when an edge is added to referencing it in the parent example


a8m (approved):

a8m:
Thanks @cliedeman 🙏

PR #100, opened 2021-07-07
Filter edges

cliedeman:
Allow skipping Edges.

Use case:
I want to implement my own resolver for an edge. For example return one 1 record from a one to many relation. 
Ent always genetes a resolve in `edge.go` that gqlgen matches. `forceResolver` does not work well in this case because it forces the generation of the model too

Depends on #99


entgql/internal/todopulid/ent/schema/verysecret.go
```diff
@@ -0,0 +1,36 @@
...
+
+// VerySecret defines the todo type schema.
+type VerySecret struct {
+	ent.Schema
+}
+
+// Mixin returns todo mixed-in schema.
+func (VerySecret) Mixin() []ent.Mixin {
+	return []ent.Mixin{
+		// "TD" declared once.
```
a8m, line 31:
```suggestion
		// "VR" declared once.
```

a8m (approved):
Looks good. You're welcome to fix the minor comment in a separate PR.

a8m:
@cliedeman, can you please rebase and rerun codegen? Thanks

PR #101, opened 2021-07-08
add ID field predicates to WhereInput

amrnt:


entgql/internal/todo/todo.graphql
```diff
@@ -90,6 +90,16 @@ input TodoWhereInput {
...
   
+  """id field predicates"""
+  id: ID
+  idNEQ: ID
+  idIn: [ID!]
+  idNotIn: [ID!]
+  idGT: ID
+  idGTE: ID
+  idLT: ID
+  idLTE: ID
```
a8m, line 101: @giautm, @cliedeman - your thoughts here, please 🙏 

We usually use the Node API for querying by ID, so I'm not convinced that it's really needed, but please share your feedback.

cliedeman: I have never found a usecase for this. I almost exclusively use the node api as the result of viewing a single objet from a url. E.g. www.example.com/person/12345

That being said. This api essentially exposes an almost complete sql query style so why would you exclude id? Maybe make it an opt in feature?

I do plan to eventually add an `Includes` annotation to pair with the `Skip` annotation and allow the user to configure if inclusion or exclusion is the default for types, edges and fields.

amrnt: An example why this would be needed: filtering a relation

```graphql
query ALL_PRODUCTS {
  products(orderBy: { direction: DESC, field: NAME }, where: { hasCategoriesWith: { id: 5 }}) {
    totalCount
    edges {
      cursor
      node {
        id
      }
    }
  }
}

```

entgql/template/where_input.tmpl
```diff
@@ -17,6 +17,7 @@ in the LICENSE file in the root directory of this source tree.
 
 {{- range $n := filterNodes $.Nodes }}
     {{ $comparableFields := list }}
+    {{ $comparableFields = append $comparableFields $n.ID }}
```
amrnt, line 20: I'm not sure if `id` should be filterable by `filterFields`

entgql/internal/todo/todo.graphql
giautm: > An example why this would be needed: filtering a relation


It make sence to me. But I think we should limit ops to: EQ/NotEQ/In/NotIn?

amrnt: makes sense! 

But it runs over the ID ops that already here*: I'm not sure how GT/GTE/... on IDs would make sense, but i mean they are available already...

*  https://github.com/ent/contrib/blob/master/entgql/internal/todo/ent/todo/where.go#L28-L108

entgql/extension.go
```diff
@@ -379,7 +379,12 @@ func (e *Extension) whereType(t *gen.Type) (string, *ast.InputObjectDefinition,
 			}),
 		}))
 	}
-	fields, err := filterFields(t.Fields)
+
+	allFields := t.Fields
+	if t.ID != nil {
+		allFields = append(allFields, t.ID)
+	}
+	fields, err := filterFields(allFields)
```
a8m, line 387: `t.ID` is always != `nil`.
```suggestion
	fields, err := filterFields(append(t.Fields, t.ID))
```

entgql/extension.go
```diff
@@ -447,6 +452,15 @@ func (e *Extension) fieldDefinition(f *gen.Field, op gen.Op) *ast.InputValueDefi
...
 	})
+
+	if f.Name == "id" {
+		def.Type = ast.NewNamed(&ast.Named{
+			Name: ast.NewName(&ast.Name{
+				Value: graphql.ID.Name(),
+			}),
+		})
+	}
+
```
a8m, line 463:
```suggestion
	if f.Name == "id" {
		def.Type = ast.NewNamed(&ast.Named{
			Name: ast.NewName(&ast.Name{
				Value: graphql.ID.Name(),
			}),
		})
	}
```

entgql/template/where_input.tmpl
a8m: We should make it possible to use the `Skip` annotation on ID field, but this can be done in future PRs.

a8m (approved):
Looks good, except for 2 minor comments. Please fix and I'll merge it. Thanks 🙏

entgql/internal/todo/todo.graphql
a8m: I suggest leaving it as it is for now and add an option to configure it using custom annotation/option in the future.

amrnt:
@a8m Fixes were pushed!

a8m:
Thanks for your contribution @amrnt ❤️

PR #105, opened 2021-07-09
Fix go vet

akfaew:
https://github.com/ent/ent/issues/1700


a8m:
@akfaew, can you please run `go generate ./...` from the `entgql` to regenerate the assets, thanks?

```
cd entgql
go generate .
go generate ./internal/...
```

akfaew:
@a8m okay, done

PR #107, opened 2021-07-11
entproto/cmd/protoc-gen-entgrpc: starting to migrate code-generation to be template based

rotemtam:
The changes made in #96 made me realize that the mixed imperative/template-based approach to code-gen in this project isn't working well enough. Code has become convoluted and hard to read (for me at least). 

This PR starts a change that will be spread across a few extra PRs to use an approach closer to the rest of the `ent` code-gen world - using `text/template`. A simple `ident` function is pushed into the template context to allow template authors to register go ident's so they are automatically qualified and imported by `protogen`'s facilities. 

The change is made gradually so the with the current PR both styles are used. 

cc @tarrencev

PR #108, opened 2021-07-11
entproto/cmd/protoc-gen-entgrpc: replace toProto func gen with template

rotemtam:
Following recently merged changes from #107, this PR replaces to `toProto` function generation with a `text/template` based approach.

The generated output has way too many new lines, I will fix this before merging.


rotemtam:
cc @tarrencev

entproto/cmd/protoc-gen-entgrpc/template/to_proto.tmpl
```diff
@@ -0,0 +1,72 @@
...
+    // toProto{{ .EntType.Name }} transforms the ent type to the pb type
+    func toProto{{ .EntType.Name }}(e *{{ .EntPackage.Ident .EntType.Name | ident }}) (*{{ .EntType.Name }}, error) {
+        v := &{{ .EntType.Name }}{}
+        {{ range .FieldMap.Fields }}
+            {{- $varName := camel .EntField.StructField -}}
+            {{- $id := print "e." .EntField.StructField -}}
+            {{ template "field_to_proto" dict "Field" . "VarName" $varName "Ident" $id }}
+            v.{{ .PbStructField }} = {{ $varName }}
+
+        {{ end }}
```
a8m, line 12:
```suggestion
        {{- range .FieldMap.Fields }}
            {{- $varName := camel .EntField.StructField -}}
            {{- $id := print "e." .EntField.StructField -}}
            {{ template "field_to_proto" dict "Field" . "VarName" $varName "Ident" $id }}
            v.{{ .PbStructField }} = {{ $varName }}
        {{- end }}
```

Please check this before landing.

entproto/cmd/protoc-gen-entgrpc/template/to_proto.tmpl
```diff
@@ -0,0 +1,72 @@
...
+                }
+            {{ else }}
+                for _, edg := range e.Edges.{{ $name }} {
+                    {{ template "field_to_proto" dict "Field" . "VarName" $varName "Ident" $id }}
+                    v.{{ $name }} = append(v.{{ $name }}, &{{ .EntEdge.Type.Name }}{
+                        {{ .EdgeIDPbStructField }}: {{ $varName }},
+                    })
+                }
+            {{ end }}
+        {{ end }}
```
a8m, line 32:
```suggestion
        {{- range .FieldMap.Edges }}
            {{- $varName := camel .EntEdge.Type.ID.StructField -}}
            {{- $id := print "edg." .EntEdge.Type.ID.StructField -}}
            {{- $name := .EntEdge.StructField -}}
            {{- if .EntEdge.Unique }}
                if edg := e.Edges.{{ $name }}; edg != nil {
                    {{ template "field_to_proto" dict "Field" . "VarName" $varName "Ident" $id }}
                    v.{{ $name }} = &{{ .EntEdge.Type.Name }}{
                        {{ .EdgeIDPbStructField }}: {{ $varName }},
                    }
                }
            {{- else }}
                for _, edg := range e.Edges.{{ $name }} {
                    {{ template "field_to_proto" dict "Field" . "VarName" $varName "Ident" $id }}
                    v.{{ $name }} = append(v.{{ $name }}, &{{ .EntEdge.Type.Name }}{
                        {{ .EdgeIDPbStructField }}: {{ $varName }},
                    })
                }
            {{- end }}
        {{- end }}
```

entproto/cmd/protoc-gen-entgrpc/template/to_proto.tmpl
```diff
@@ -0,0 +1,72 @@
...
+        {{ .VarName }}Value, err := {{$id}}.Value()
+        if err != nil {
+            return nil, err
+        }
+        {{ .VarName }}Typed, ok := {{ .VarName }}Value.({{ $conv.ToProtoValuer }})
+        if !ok {
+            return nil, {{ qualify "errors" "New" }}("casting value to {{ $conv.ToProtoValuer }}")
+        }
+        {{ .VarName}} := {{ ident $conv.ToProtoConstructor }}({{ .VarName}}Typed)
+    {{- else if $conv.ToProtoValuer -}}
```
a8m, line 58:
```suggestion
    {{- if $conv.ToProtoConversion }}
        {{- $id = print $conv.ToProtoConversion "(" $id ")" }}
    {{- end -}}
    {{- if $conv.ToEntMarshallerConstructor.GoName }}
        {{ .VarName }}, err := {{ $id }}.MarshalBinary()
        if err != nil {
            return nil, err
        }
    {{- else if and $conv.ToProtoValuer $conv.ToProtoConstructor.GoName }}
        {{ .VarName }}Value, err := {{$id}}.Value()
        if err != nil {
            return nil, err
        }
        {{ .VarName }}Typed, ok := {{ .VarName }}Value.({{ $conv.ToProtoValuer }})
        if !ok {
            return nil, {{ qualify "errors" "New" }}("casting value to {{ $conv.ToProtoValuer }}")
        }
        {{ .VarName}} := {{ ident $conv.ToProtoConstructor }}({{ .VarName}}Typed)
    {{- else if $conv.ToProtoValuer }}
```

entproto/cmd/protoc-gen-entgrpc/template/to_proto.tmpl
```diff
@@ -0,0 +1,72 @@
...
+        }
+        {{ .VarName }}, ok := {{ .VarName }}Value.({{ $conv.ToProtoValuer }})
+        if !ok {
+            return nil, {{ qualify "errors" "New" }}("casting value to {{ $conv.ToProtoValuer }}")
+        }
+    {{- else if $conv.ToProtoConstructor.GoName -}}
+        {{ .VarName }} := {{ ident $conv.ToProtoConstructor }}({{ $id }})
+    {{- else -}}
+        {{ .VarName }} := {{ $id }}
+    {{- end -}}
```
a8m, line 71:
```suggestion
    {{- else if $conv.ToProtoConstructor.GoName }}
        {{ .VarName }} := {{ ident $conv.ToProtoConstructor }}({{ $id }})
    {{- else }}
        {{ .VarName }} := {{ $id }}
    {{- end }}
```

a8m (approved):

tarrencev:
One issue i encountered, that I'll mention here but we can solve later, is ent entity names can conflict with go reserved names. For example, an entity can be called `for`. We should add a prefix/suffix to generated var names to avoid these collisions.

a8m:
> One issue i encountered, that I'll mention here but we can solve later, is ent entity names can conflict with go reserved names. For example, an entity can be called `for`. We should add a prefix/suffix to generated var names to avoid these collisions.

Can you share an example, because afaik, it’s not possible to name schema types this way:
https://github.com/ent/ent/blob/master/entc/gen/type.go#L796

tarrencev:
> > One issue i encountered, that I'll mention here but we can solve later, is ent entity names can conflict with go reserved names. For example, an entity can be called `for`. We should add a prefix/suffix to generated var names to avoid these collisions.
> 
> Can you share an example? Because afaik, it’s not possible to name schema types this way:
> https://github.com/ent/ent/blob/master/entc/gen/type.go#L796

I have a schema that compiles with
```
field.Int("for").
			GoType(BigInt{}).
			SchemaType(BigIntSchemaType).
			DefaultFunc(func() BigInt {
				return NewBigInt(0)
			}),
```

a8m:
> > > One issue i encountered, that I'll mention here but we can solve later, is ent entity names can conflict with go reserved names. For example, an entity can be called `for`. We should add a prefix/suffix to generated var names to avoid these collisions.
> > 
> > 
> > Can you share an example? Because afaik, it’s not possible to name schema types this way:
> > https://github.com/ent/ent/blob/master/entc/gen/type.go#L796
> 
> Hmm interesting. I have a schema that compiles with
> 
> ```
> field.Int("for").
>   GoType(BigInt{}).
>   SchemaType(BigIntSchemaType).
>   DefaultFunc(func() BigInt {
>     return NewBigInt(0)
>   }),
> ```

You're right. That's expected. I thought you meant the entity-type, but fields can be named as Go keywords.

rotemtam:
> One issue i encountered, that I'll mention here but we can solve later, is ent entity names can conflict with go reserved names. For example, an entity can be called `for`. We should add a prefix/suffix to generated var names to avoid these collisions.

@tarrencev , did you encounter this issue in relation to this PR or is this a general comment?

tarrencev:
> > One issue i encountered, that I'll mention here but we can solve later, is ent entity names can conflict with go reserved names. For example, an entity can be called `for`. We should add a prefix/suffix to generated var names to avoid these collisions.
> 
> 
> 
> @tarrencev , did you encounter this issue in relation to this PR or is this a general comment? 

Sorry i was unclear. This is an issue that was introduced by the value scanner work

PR #110, opened 2021-07-14
entgql: add initial support for schema generation

jlarfors:
This is a WIP hence the draft PR. Looking for feedback whilst testing this in my project locally.

Feedback welcome, especially on the configuration options/assumptions that are made about what to generate. This is quite opinionated and not that customisable right now.


cliedeman:
Hi @jlarfors thanks you for the contribution. 

Can you explain the use case for this? 

How does it relate to the existing pagination templates?

Ciaran

jlarfors:
> Hi @jlarfors thanks for the contribution.
> 
> Can you explain the use case for this?
> 
> How does it relate to the existing pagination templates?
> 
> Ciaran

Hi @cliedeman , the use case is to generate the GraphQL schema types and Query object. There is an option to support both the relay specification (using the existing pagination templates) and a "vanilla" graphql interface (for which I added a `filter_input.tmpl` to achieve similar the same filters as pagination, without the pagination :) ).

cliedeman:
Hi @jlarfors thanks for explaining.

I think should rather be a private template and not inside the contrib repository. The reason being filtering styles will vary greatly when not using the relay connection spec. 

Ciaran

jlarfors:
@cliedeman sure, that's fine and I can remove the filter template from contrib.

What about the graphql schema generation (which is the main work in this PR)?

jlarfors:
`filter_input.tmpl` has now been removed so the only change in this PR is the one regarding graphql schema generation, which is the part I am actually interested :)

cliedeman:
@jlarfors apologies I missed that file on my phone because it was supressed 🙈.

I am a big fan of this type of integration - https://github.com/ent/contrib/pull/78. This is close to the style @a8m ending up going with so I will review it when I can 👍

entgql/extension.go
```diff
@@ -136,12 +141,57 @@ func WithTemplates(templates ...*gen.Template) ExtensionOption {
 	}
 }
 
+// WithNaming takes either "snake" or "camel" and sets that as the naming
+// convention when generating the GraphQL schema
+func WithNaming(s string) ExtensionOption {
```
cliedeman, line 146: This is quite odd. I have encountered a snake_case graphql schema. Is this a use case you specifiically need?

entgql/extension.go
```diff
@@ -215,7 +269,7 @@ func (e *Extension) mapScalar(f *gen.Field, op gen.Op) string {
 		scalar = graphql.Boolean.Name()
 	case f.IsEdgeField():
 		scalar = graphql.ID.Name()
-	case t.Numeric():
+	case t.Integer():
```
cliedeman, line 272: Nice catch, This is a pretty nasty bug you should probably create a seperate pr for

entgql/extension.go
```diff
@@ -231,6 +285,299 @@ func (e *Extension) mapScalar(f *gen.Field, op gen.Op) string {
...
+			nodes, err := filterNodes(g.Nodes)
+			if err != nil {
+				return err
+			}
+			if err := next.Generate(g); err != nil {
+				return err
+			}
+
+			addTimeScalarDefinition(scalars)
+			addNodeDefinition(interfaces)
```
cliedeman, line 309: this is part of the relay specification

entgql/extension.go
```diff
@@ -231,6 +285,299 @@ func (e *Extension) mapScalar(f *gen.Field, op gen.Op) string {
...
+					}),
+				},
+			},
+		)
+	}
+	return hasOrderByType
+}
+
+// addOrderByDefinitions creates a GraphQL query field for the given node, adding
+// the relevanant arguments based on whether the relay spec should be followed,
```
cliedeman, line 477: typo relevanant

entgql/extension.go
```diff
@@ -231,6 +285,299 @@ func (e *Extension) mapScalar(f *gen.Field, op gen.Op) string {
...
+	for _, f := range node.Fields {
+		obj.Fields = append(obj.Fields, ast.NewFieldDefinition(
+			&ast.FieldDefinition{
+				Name: newASTName(f.Name),
+				Type: newASTNamed(e.mapScalar(f, gen.EQ)),
+			}),
+		)
+	}
+
+	for _, edge := range node.Edges {
```
cliedeman, line 381: we will need an annotation here. I don't think edges should be included by default.

Also some edges you will want to be connections and some arrays

cliedeman (changes requested):

entgql/extension.go
```diff
@@ -231,6 +285,299 @@ func (e *Extension) mapScalar(f *gen.Field, op gen.Op) string {
...
+				if e.whereFilters {
+					// Create the where input types
+					err := e.addWhereType(inputs, node)
+					if err != nil {
+						return err
+					}
+				}
+
+				if e.relaySpec {
+					e.addRelayObjectDefinitions(objects, node)
```
cliedeman, line 330: This should be driven by an annotation

entgql/extension.go
```diff
@@ -231,6 +285,299 @@ func (e *Extension) mapScalar(f *gen.Field, op gen.Op) string {
...
+					e.addQueryDefinition(queries, node, true, hasOrderByType)
```
cliedeman, line 331: This you should probably leave up to the user to define. For example If they use custom filters. Or make it an opt in annotation

entgql/extension.go
```diff
@@ -231,6 +285,299 @@ func (e *Extension) mapScalar(f *gen.Field, op gen.Op) string {
...
+		ast.NewFieldDefinition(&ast.FieldDefinition{
+			Name: newASTName(e.naming("total_count")),
+			Type: newASTNamed(graphql.Int.String() + "!"),
+		}),
+		ast.NewFieldDefinition(&ast.FieldDefinition{
+			Name: newASTName(e.naming("page_info")),
+			Type: newASTNamed("PageInfo!"),
+		}),
+		ast.NewFieldDefinition(&ast.FieldDefinition{
+			Name: newASTName("edges"),
```
cliedeman, line 554: I prefer to have the option to make these more strict.
E.g.
make the `edges` and `node` not null. What do you think?

cliedeman:
@jlarfors the pr is quite big.

Can you create a pr first with these refactorings and the bug?, that should make it easier to review

```go
func newASTName(name string) *ast.Name {
	return ast.NewName(&ast.Name{Value: name})
}

func newASTNamed(name string) *ast.Named {
	return ast.NewNamed(&ast.Named{
		Name: newASTName(name),
	})
}
```

a8m (commented):
Amazing! 🚀 
Thanks for your contribution @jlarfors, and sorry for the delay. 

Can you please break it into smaller PRs? Ping me on our Slack channel if you need help with that.

jlarfors:
Hi @cliedeman and @a8m and sorry for the late reply! I was on holiday for a bit, but back now.
I also did not see that there was already a considerable contribution to the graphql code generation, with annotations. I haven't reviewed it in detail yet but it'd be great to combine these IMO.

I should get some time this week to look into this and get some smaller PRs together... So please stand by

entgql/template/filter_input.tmpl
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+    "context"
+
+	{{- range $n := $gqlNodes }}
+		"{{ $.Config.Package }}/{{ $n.Package }}"
+	{{- end }}
+)
+
+{{- range $n := filterNodes $.Nodes }}
```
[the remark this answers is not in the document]
jlarfors: Thanks! This has been moved to private code now but i fixed it there :)

entgql/extension.go
jlarfors: In our case, the schema is something that the users will interact with directly, in various different ways and so far it is `snake_case` everywhere... and whilst this is not "standard" for Graphql I think it would be nice to support.
So yes, it's a case we would really like

entgql/extension.go
jlarfors: Is this a question or a comment? The node definitions are always needed, whether using relay spec or not. The relay spec would simply wrap it into a `Node` rather than returning the type directly.

entgql/extension.go
jlarfors: Agree. I had not thought about making an edge an array... and actually I can see some nice use cases for it !

entgql/extension.go
jlarfors: Makes sense and can be done!

entgql/extension.go
cliedeman: My Mistake. I confused this with the `node(id: ID!)` query

entgql/extension.go
yonidavidson: I also bumped into this a few days ago, https://github.com/ent/contrib/pull/114/files

giautm:
Hello, I'm going to close this pull request. Because almost feature you are committed to is ready on the master branch.

Thank you for your contribution, and get started on these features. Best regards,

lrstanley:
@giautm is there somewhere I could look at to determine the status of/track the higher level functionality?

giautm:
You can follow this issue https://github.com/ent/ent/issues/2351, we will send updates to it.

PR #111, opened 2021-07-20
entgql: support skipping id fields in WhereInput

a8m:


entgql/template/where_input.tmpl
```diff
@@ -15,9 +15,13 @@ in the LICENSE file in the root directory of this source tree.
 
 {{ template "import" $ }}
 
-{{- range $n := filterNodes $.Nodes }}
-    {{ $comparableFields := list }}
-    {{ $comparableFields = append $comparableFields $n.ID }}
+{{ range $n := filterNodes $.Nodes }}
+    {{ $comparableFields := list $n.ID }}
```
cliedeman, line 19: can we not do something like `filterNodes (list $n.ID)` to future proof adding more logic

a8m: Do you mean `filterFields`? The only problem is that `list` returns `[]interface{}` and `filterFields` accepts `[]*gen.Field` as an input.

cliedeman: ah thats unfortunate. I will try to address it when I add the Includes Annotation

PR #113, opened 2021-08-09
Adding entprom  - a prometheus hook

yonidavidson:
Adds a hook with Readme and examples for using Prometheus hook when mutating data on the graph.


entprom/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Attach hook only for specific schema - for example User
+
+```go
+// Hooks of the User.
+func (User) Hooks() []ent.Hook {
+    return []ent.Hook{
+        entprom.Hook(),
+    }
+}
```
rotemtam, line 66: @yonidavidson, will this work if you rn the `entprom` hook on multiple, different schemas?

entprom/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Next, wherever you are setting up your HTTP listener:
+```go
+//get from github.com/prometheus/client_golang/prometheus/promhttp
+http.Handle("/metrics", promhttp.Handler())
+```
+
+After your server is running and a few mutations were made, you can get the metrics from the `/metrics` endpoint.
+
+TODO(yoni): add curl example and sample output.
```
rotemtam, line 34: please

entprom/go.mod
```diff
@@ -0,0 +1,10 @@
+module entprom
```
rotemtam, line 1: remove this file, use the global one with everyone else. we will split the modules if needed at a later point in time

entprom/README.md
yonidavidson: only if there is a difference in one of the labels.

giautm:
Why not opentelemetry?

yonidavidson:
> Why not opentelemetry?

Ideally we would want to integrate with the OpenTelemetry project, but it appears that the go metrics integration is on hold there:
[1]: The development of the metrics API and SDK has paused due to limited development resources, prioritization of a stable Traces release, and instability of the official overall design from the OpenTelemetry specification. Pull Requests for metrics-related issues are not being accepted currently outside of security vulnerability mitigations.

rotemtam:
I agree with @yonidavidson here. It appears that OpenTelemetry _should_ be the future, but for now, it's on hold and we'd rather have ent users something that's stable.

giautm:
I knew OpenTelemetry was not complete yet, but don't know they pause development on it, sad.

But, [opencensus](https://opencensus.io/stats/) is still here for production usage. Prometheus looks like vendor-lock, If I want to use this hook on GCP/AWS then I have to re-write.

PS: I still usage prometheus on our system, but with opencensus integration.

cliedeman:
Would it be feasible to create a entmetrics package that has different templates for each implementation. I know tracing changes quite a bit for opentelemetry but with metrics it should be a lot more straightforward?

yonidavidson:
> Would it be feasible to create an entmetrics package that has different templates for each implementation. I know tracing changes quite a bit for opentelemetry but with metrics it should be a lot more straightforward?

It's an interesting  idea,  If you can write a few words about it in a new issue I'll be able to review.

rotemtam:
hey @cliedeman @giautm 

thanks for the great feedback. I completely agree that ent should provide a robust solution that is decoupled from any specific metric collection backend. Our users are most likely using many different metric collection backends such as Prometheus, InfluxDB, CloudWatch, datadog, etc. 

This issue is very common and widespread and has spawned initiatives such as [OpenCensus](https://opencensus.io/) and [OpenTelemetry](https://opentelemetry.io/). I am confident that this problem should and will be solved in a standards-based layer where the entire industry will collaborate to decouple the interface and implementation of such systems. Similar things are happening with the [Open Container Initiative](https://opencontainers.org/), [Service Mesh Initiative](https://smi-spec.io/).  Currently, it appears that the industry will converge around [OpenTelemetry](https://opentelemetry.io/) but as @yonidavidson mentioned, support for metrics+go there seems to have been put in the icebox. 

I have big hopes for the standard-based approach and I am following the development of OpenTelemetry. As such, I don't think that `ent` should develop an "in-house" abstraction for metric backends. 

The way I think about this PR is that it offers support for a basic, but well-defined use case, to get something working "out the door" for feedback.  Just as we released `entproto` in a fairly limited scope and iterated over a few months to bring it to the point where it's actually usable, I think this PR is just the first in an initiative to bring proper instrumentation to Ent. 

Consider, for example, that we don't have support for "interceptor" like API yet, which would be required to properly instrument reads from the graph (Get/Update). In other words, this is a _very_ initial step towards something which offers broad support for multiple metric collection backends. 

If we don't see any progress in OpenTelemetry within the next few months towards making metrics generally available in Go, we should definitely consider going with @cliedeman's approach to building some local interface that abstracts the metric collection backend. This interface would have a small subset of the OpenTelemetry API surface limited to the things ent needs. 

I hope you agree that releasing `entprom` is a step in the right direction

yonidavidson:
Closing for discussion.

PR #117, opened 2021-08-17
all: enable dependabot

masseelch:
Main repo already uses it, why does this one not? ent has had two releases already.


a8m:
Thanks @masseelch 🙏🏻

PR #124, opened 2021-08-18
support gql generated prefix (#6)

yonidavidson:
added gql_ prefix to all files generated.
Create a small generation hook to migrate the codebase to the new hook.
Validate that existing package name is not deleted

the error generate: open PROJECT_NAME/ent/node.go: no such file or directory displays when file names are changed (old deleted new are created).


go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/mattn/go-sqlite3 v1.14.7
 	github.com/mitchellh/mapstructure v1.4.1
 	github.com/oklog/ulid/v2 v2.0.2
+	github.com/prometheus/common v0.4.0
```
a8m, line 26: ??

a8m:
Thanks for your contribution @yonidavidson. Merged with #125

PR #135, opened 2021-09-02
fix #1882

tsingsun:
fix #1882 field with underscore and number generate bug in protoc-gen-entgrpc


entproto/fieldmap.go
```diff
@@ -176,3 +175,60 @@ func extractEntEdgeByName(entType *gen.Type, name string) (*gen.Edge, error) {
...
+
+// CamelCase returns the CamelCased name.
+// If there is an interior underscore followed by a lower case letter,
+// drop the underscore and convert the letter to upper case.
+// There is a remote possibility of this rewrite causing a name collision,
+// but it's so remote we're prepared to pretend it's nonexistent - since the
+// C++ generator lowercases names, it's extremely unlikely to have two fields
+// with different capitalizations.
+// In short, _my_field_name_2 becomes XMyFieldName_2.
+func CamelCase(s string) string {
```
a8m, line 197: Can we make this function private, and mention that it was copied from https://github.com/golang/protobuf/blob/v1.5.2/protoc-gen-go/generator/generator.go#L2648 ?

a8m:
Thanks for the contribution @tsingsun. Please fix the minor comment above, and connect your Git email (on your computer) to the GitHub settings if you want this commit to be listed on your account.

image

entproto/fieldmap.go
tsingsun: done

rotemtam:
hey @tsingsun,
thanks a lot for working on this. 
please add this scenario to a unit test so we can verify it works? here are some details on how I work on this project

### The dev loop
The tests for protoc-gen-entgrpc are in entproto/internal/todo. the dev flow should be:

* Re-build and install the protoc plugin
* Regenerate the ent project (which invokes entproto to generate proto files and regenerates the service definition and implementation)
* Run the unit tests (these are unit tests against the generated code)

* When working on protoc-gen-entgrpc, I usually run this command:

```
go get entgo.io/contrib/entproto/cmd/protoc-gen-entgrpc && go generate ./internal/todo/... && go test ./internal/todo/ent/proto/...
```
As part of this PR we should add to one of the schemas entproto/internal/todo (for example, https://github.com/ent/contrib/blob/master/entproto/internal/todo/ent/schema/attachment.go) a field which would represent the use case - for example, a string named "a_b_3", do the same with an edge. 

The code generation is a bit sensitive to protoc and plugin versions, if you want to run the generation command within Docker you can use the image described in [this discussion](https://github.com/ent/ent/discussions/1676)

rotemtam (changes requested):
looks good, please add tests (see comment).

tsingsun:
has added tests and found the same problem in to_proto.tmpl

rotemtam (approved):

PR #139, opened 2021-09-14
schemast: support field.Time

rotemtam:
cc @zeevmoney

### DIFF
--- a/schemast/field.go
+++ b/schemast/field.go
@@ -15,10 +15,12 @@
 package schemast
 
 import (
+	"errors"
 	"fmt"
 	"go/ast"
 	"go/token"
 	"reflect"
+	"runtime"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


schemast/field.go
```diff
@@ -197,6 +199,13 @@ func defaultExpr(d interface{}) (ast.Expr, error) {
 			Value: strconv.FormatBool(d.(bool)),
 		}
 		return lit, nil
+	case reflect.Func:
+		f := runtime.FuncForPC(v.Pointer()).Name()
+		parts := strings.Split(f, ".")
+		if len(parts) != 2 {
+			return nil, errors.New("schemast: only selector exprs are supported for default func")
+		}
```
a8m, line 207: Please explain why?

rotemtam: consider that you have a field
```go
field.Time("time").Default(func() time.Time { /.../})
```
At runtime, the field descriptor holds a reference to a function, it cannot build that functions AST.
If that function is declared in an external package, and referenced by a selector expression, for example `time.Now`, we can retain that reference and things should work fine when rewriting (because running `goimports` can fix the import for the user).

rotemtam: We can also support `field.Time` and not support `Default(func)` at all

a8m: We can support this by using pointers to the AST file and copy the string content as is, right?

a8m (approved):
Looks good 🚀

PR #155, opened 2021-11-04
Implement Optional GetAll Method for Services

JeremyV2014:
Addresses https://github.com/ent/ent/issues/2107

ToDo:

- [ ] Implement unit tests for validation

### DIFF
--- a/entproto/service.go
+++ b/entproto/service.go
@@ -39,19 +39,43 @@ var (
 	errNoServiceDef = errors.New("entproto: annotation entproto.Service missing")
 )
 
+type ServiceOptionsEnum int
+
+const (
+	// OptionGenGetAll generates a GetAll RPC for retrieving all ids of existing entities.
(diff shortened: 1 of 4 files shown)


entproto/service.go
```diff
@@ -39,19 +39,43 @@ var (
...
-func Service() schema.Annotation {
-	return service{Generate: true}
+func Service(options ...ServiceOptionsEnum) schema.Annotation {
+	newService := &service{
+		Generate: true,
+		Options:  &serviceOptions{},
+	}
+
+	for _, option := range options {
+		switch option {
```
JeremyV2014, line 69: Linter is failing me for this switch case because it only has one case. The idea behind the switch case is that the options are extendable beyond this.

rotemtam:
Hey @JeremyV2014 as I responded on the issue, really well done on the detailed work on this, but I have some reservations about adding this to the package, let's continue the discussion there. Closing for now, once we nail the design you can pick up from here

PR #157, opened 2021-11-09
Support string slice in entproto

jordanrule:
Issue: https://github.com/ent/ent/issues/1585

✅ `golangci-lint run` passes in docker container
✅ `entproto/go generate ./...` generates no diffs
✅ `entproto/go test ./...` passes
✅ added appropriate unit tests to `entproto/internal/`

Verified working by creating a sample grpc service from a `field.Strings`, creating a sample user and querying it:

```
go run ./cmd/client
2021/11/09 15:01:48 user created with id: 1
2021/11/09 15:01:48 retrieved user with id=1: id:1 name:"user_8308003752681997053" email_address:"user_6125529902775396743@example.com" my_strings:"one" my_strings:"two" my_strings:"three"
```


rotemtam:
Hi,
Thanks for working on this! Please add unit tests? Lmk if you need help
Thanks

jordanrule:
I'm looking at the linter failures, I hadn't noticed the specific configuration in `.golangci`.  For the tests and generate, I used `replace entgo.io/contrib/entproto => ../contrib/entproto` in my `go.mod`, I'm not sure how to handle this change for the PR builder.

rotemtam:
> I'm looking at the linter failures, I hadn't noticed the specific configuration in `.golangci`. For the tests and generate, I used `replace entgo.io/contrib/entproto => ../contrib/entproto` in my `go.mod`, I'm not sure how to handle this change for the PR builder.

Hey
I've added more specific guidelines on how to contribute to `entproto`:
https://github.com/ent/contrib/tree/master/entproto#contributing

jordanrule:
Thanks for the container and specific build, I'm fairly confident this will now pass the builder.  Do note the constructors added in `entproto/slice`, I'm not sure about naming or if there is a cleaner approach.

entproto/internal/todo/ent/proto/entpb/entpb.proto
```diff
@@ -132,19 +132,23 @@ message User {
 
   google.protobuf.BoolValue opt_bool = 15;
 
-  google.protobuf.StringValue big_int = 17;
+  repeated google.protobuf.StringValue opt_strings = 16;
```
rotemtam, line 135: why is field 17 deleted?

entproto/internal/todo/ent/proto/entpb/entpb.proto
```diff
@@ -132,19 +132,23 @@ message User {
...
 
-  google.protobuf.Int32Value b_user_1 = 18;
+  google.protobuf.StringValue big_int = 18;
 
-  float height_in_cm = 19;
+  google.protobuf.Int32Value b_user_1 = 19;
```
rotemtam, line 139: why is field 19 changing?

entproto/internal/todo/ent/schema/user.go
```diff
@@ -87,24 +87,29 @@ func (User) Fields() []ent.Field {
 		field.Bool("opt_bool").
 			Optional().
 			Annotations(entproto.Field(15)),
+		field.Strings("opt_strings").
```
rotemtam, line 90: with protos, you always add new fields with new tag numbers, never replace existing ones. lets do that in the tests as well

rotemtam (changes requested):
See some more comments.

Something that occured to me and I want to discuss with you. 

when someone says `field.Stings().Optional` it does not mean that they expect a list of nillable strings (i.e an array like ["a",null,"b"]) which is what you did by creating a list of wrapperspb.StringValue. Right?

 It should mean that they expect a list of string literals, or nothing. (i.e the db value can be nil).

WDYT?

jordanrule:
I'm not sure syntactically there is a difference given my understanding is the wrappers are used to conform to `omitempty` (`nil` would be disallowed in a []string per the setter), but I agree it makes more sense semantically to represent as a `ListValue`.  This latest commit utilizes `repeated string` as the required type, and `ListValue` as the optional type.

rotemtam:
Hey @jordanrule 
thanks again for working on. sorry, i wasn't clear enough on my previous message. 
I'm not 100% sure about how to represent `field("field").Strings().Optional()`, current solution is an option but i'd like to give it more thought. 

My suggestion is to remove support for optional for now, and discuss how optional would look in a subsequent PR.
we could do it on this PR, but it in my experience its better to ship what's ready and focus more on what we're not sure about. I hope this makes sense

rotemtam:
@jordanrule closing this due to no progress in past 2 months . If you want to reboot this ping me here or on Ent discord 🙏

PR #158, opened 2021-11-11
Implement Service Method Generation Control

JeremyV2014:
Allow users to specify which methods they would like to generate for a gRPC service.


rotemtam:
Hey @JeremyV2014,
Thanks for working on this!

A few things:

Define the method enums with`1 << iota` like in https://github.com/ent/ent/blob/3ba2b4e1737a4db1cb7ee9b81c7bff5961f515fa/ent.go#L354-L360
For example, https://play.golang.org/p/HiXdR5Ck1Ni

How about making the API:

```go
// only render Get and Delete 
entproto.Service(
  entproto.Methods(entproto.MethodGet | entproto.MethodDelete)
)

// render all
entproto.Service(
  entproto.Methods(entproto.MethodAll)
)

// default behavior equivalent to MethodAll
// to make this work you will need to use the functional options pattern
entproto.Service()
```

entproto/internal/entprototest/ent/schema/blog_post.go
```diff
@@ -52,6 +52,8 @@ func (BlogPost) Fields() []ent.Field {
 func (BlogPost) Annotations() []schema.Annotation {
 	return []schema.Annotation{
 		entproto.Message(),
-		entproto.Service(),
+		entproto.Service(
+			entproto.Methods(
+				entproto.MethodAll)),
```
a8m, line 57:
```suggestion
				entproto.MethodAll,
			),
		),
```

entproto/service.go
```diff
@@ -25,33 +25,52 @@ import (
...
 	ServiceAnnotation        = "ProtoService"
-	get               method = "Get"
-	create                   = "Create"
-	update                   = "Update"
-	delete_                  = "Delete"
+	MethodAll         Method = "All"
+	MethodCreate      Method = "Create"
+	MethodGet         Method = "Get"
+	MethodUpdate      Method = "Update"
+	MethodDelete      Method = "Delete"
```
a8m, line 34: Maybe use bit flags instead?

```go
// A Method defines ...
type Method uint

const (
	MethodGet Method = 1 << iota
	MethodCreate
	MethodUpdate
	MethodDelete
)

// Is reports whether m is match the given method.
func (m Method) Is(n Method) bool { return m&n != 0 }

var MethodAll = MethodGet | MethodCreate | MethodUpdate | MethodDelete
```

entproto/service.go
```diff
@@ -25,33 +25,54 @@ import (
...
+	Methods  Method
 }
 
 func (service) Name() string {
 	return ServiceAnnotation
 }
 
-func Service() schema.Annotation {
-	return service{Generate: true}
+func Service(methods ...Method) schema.Annotation {
```
rotemtam, line 59: should be a functional option. think that we may want other parameters. in other orders
`entproto.Service(entproto.MethodAll)` should not work. 

if you're not familiar with the concept see this post its pretty cool:
https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

and see the implementation in a file right next to you: https://github.com/ent/contrib/blob/master/entproto/message.go#L28-L40

JeremyV2014: Thanks! I'll work on incorporating this into my PR. I also didn't realize I could do bit shifts with iota. Very cool!

If it wasn't clear, I'm still relatively new to Go; I'm still learning some of the design patterns for it.

rotemtam: You're doing great! Thanks for being patient and pushing this through!

JeremyV2014: Okay, I think I've implemented this design pattern appropriately now. It is definitely quite neat!

rotemtam:
The implementation looks right, thanks! Please add tests?

JeremyV2014:
> Please add tests?

To internal/entprototest/service_test.go?

entproto/internal/entprototest/service_test.go
```diff
@@ -14,30 +14,140 @@
...
+		svc := fd.FindService(fmt.Sprintf("entpb.%sService", tc.schemaName))
+		suite.NotNil(svc)
+
+		for _, m := range tc.includedMethods {
+			getMeth := svc.FindMethodByName(m.name)
+			suite.Require().NotNil(getMeth)
+
+			if m.formatInput {
+				m.input = m.getFormatInput(tc.schemaName)
+			}
```
rotemtam, line 138: sorry i dont understand, what is this doing? seems like formatInput is always true?

entproto/internal/entprototest/service_test.go
```diff
@@ -14,30 +14,140 @@
...
+		input:        "Delete%sRequest",
+		output:       "Empty",
+		formatInput:  true,
+		formatOutput: false,
+	}
+)
+
 func (suite *AdapterTestSuite) TestServiceGeneration() {
-	fd, err := suite.adapter.GetFileDescriptor("BlogPost")
-	suite.Require().NoError(err)
```
rotemtam, line 19: This test code gets a little too clever for what it needs to do.  I had to read it a few times to understand why you need to formatInput/formatOutput, etc. 

That's just too much logic. Tests should be self explanatory, very _very_ simple (since they aren't tested themselves). DRY is not the most important design principle. Copy-paste is perfectly fine.

I wrote about this topic in my personal blog some time ago:
https://rotemtam.com/2020/05/18/the-dry-principle-is-bad-advice/

I would go back to just having short assertions about what you're verifying.

entproto/internal/todo/ent/schema/user.go
```diff
@@ -38,7 +38,14 @@ type User struct {
...
 		entproto.Message(),
-		entproto.Service(),
+		entproto.Service(
+			entproto.Methods(
+				entproto.MethodCreate |
+					entproto.MethodGet |
+					entproto.MethodUpdate |
+					entproto.MethodDelete,
+			),
+		),
```
rotemtam, line 48: I would leave this as is to prove that its not broken by your change

entproto/service.go
```diff
@@ -25,33 +25,58 @@ import (
...
 var (
 	errNoServiceDef = errors.New("entproto: annotation entproto.Service missing")
 )
 
+type Method uint
+
+// Is reports whether method m matches given method n.
+func (m Method) Is(n Method) bool { return m&n != 0 }
+
+func Methods(methods Method) ServiceOption {
```
rotemtam, line 46: godoc

rotemtam (commented):
Added some final comments.
Also if you can, please update the README file for entproto (in a separate PR is fine) and update the gRPC documentation on the ent site (also in a separate PR).

rotemtam (changes requested):
Also notice that CI is failing

JeremyV2014:
> Also notice that CI is failing

Looks like I'm missing copyright headers in the new files.

entproto/internal/entprototest/service_test.go
JeremyV2014: The idea was to leave this here for future methods that might have an input message that isn't based on the schema name. (Like google.protobuf.Empty)

JeremyV2014:
> Also if you can, please update the README file for entproto (in a separate PR is fine)

Should entproto.Service() get a higher header level? It seems like this is quite a bit of information to add to that section.

> update the gRPC documentation on the ent site (also in a separate PR).

How would you like this introduced on the website? Are you thinking of a new page discussing method generation options for services? I could walk through modifying the User service to generate only Get and Create methods. This could become the last page for the gRPC tutorial. It could be called "Service Generation Options" or something like that.

entproto/internal/entprototest/ent/schema/all_methods_service.go
```diff
@@ -0,0 +1,47 @@
...
+// Fields of the AllMethodsService.
+func (AllMethodsService) Fields() []ent.Field {
+	return nil
+}
+
+// Edges of the AllMethodsService.
+func (AllMethodsService) Edges() []ent.Edge {
+	return nil
+}
+
```
a8m, line 37:
```suggestion
```

You can delete these methods for empty fields/edges/indexes/etc.

entproto/internal/entprototest/ent/schema/one_method_service.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Fields of the OneMethodService.
+func (OneMethodService) Fields() []ent.Field {
+	return nil
+}
+
+// Edges of the OneMethodService.
+func (OneMethodService) Edges() []ent.Edge {
+	return nil
+}
```
a8m, line 36:
```suggestion
```

You can delete these methods for empty fields/edges/indexes/etc.

entproto/internal/entprototest/ent/schema/two_method_service.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Fields of the TwoMethodService.
+func (TwoMethodService) Fields() []ent.Field {
+	return nil
+}
+
+// Edges of the TwoMethodService.
+func (TwoMethodService) Edges() []ent.Edge {
+	return nil
+}
```
a8m, line 36:
```suggestion
```

ditto

entproto/service.go
```diff
@@ -25,33 +25,59 @@ import (
...
+
+// Service annotates an ent.Schema to specify that protobuf service generation is required for it.
+func Service(opts ...ServiceOption) schema.Annotation {
+	s := service{
+		Generate: true,
+	}
+	for _, apply := range opts {
+		apply(&s)
+	}
+	// Default to generating all methods
```
a8m, line 73:
```suggestion
	// Default to generating all methods.
```

a8m (approved):
Looks good to me.

rotemtam (approved):
LGTM
Great job @JeremyV2014, thanks for working on this!

PR #159, opened 2021-11-11
entproto: documenting local dev flow

rotemtam:


entproto/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.17.3
```
a8m, line 1: Why not `latest`? 

```suggestion
FROM golang
```

a8m (approved):

entproto/Dockerfile
rotemtam: I prefer being specific to make this as deterministic as i can.

PR #162, opened 2021-11-16
entproto: implement list service method

JeremyV2014:
Addressing https://github.com/ent/ent/issues/2107


entproto/internal/todo/ent/proto/entpb/entpb_user_service.go
```diff
@@ -342,3 +346,87 @@ func (svc *UserService) Delete(ctx context.Context, req *DeleteUserRequest) (*em
...
+	case pageSize == 0 || pageSize > entproto.MaxPageSize:
+		pageSize = entproto.MaxPageSize
+	}
+	listQuery := svc.client.User.Query().
+		Order(ent.Desc(user.FieldID)).
+		Limit(pageSize + 1)
+	if req.GetPageToken() != nil {
+		buf := bytes.Buffer{}
+		bytes, err := base64.StdEncoding.DecodeString(req.PageToken.GetValue())
+		if err != nil {
```
rotemtam, line 373: the code for encoding / decoding the token seems like it can be simplified. 
why do we need so many layers (a local token struct, a byte buffer, gob, etc...)
our goal is to have a base64 representation of the id of the last element we read, right? can this be done with less?

rotemtam (changes requested):

JeremyV2014:
@rotemtam ,

[AIP-132](https://google.aip.dev/132) states:

> The comment above the page_size field should document the maximum allowed value, as well as the default value if the field is omitted (or set to 0).

I was trying to determine how I could attach comments to the proto fields through the FieldDescriptorProto with no success. Do you know of a way to do this?

rotemtam:
> @rotemtam ,
> 
> [AIP-132](https://google.aip.dev/132) states:
> 
> > The comment above the page_size field should document the maximum allowed value, as well as the default value if the field is omitted (or set to 0).
> 
> I was trying to determine how I could attach comments to the proto fields through the FieldDescriptorProto with no success. Do you know of a way to do this?

i think the pb lib we're using has this concept:
https://github.com/ent/contrib/blob/712b44424d50a097eea13d843148a7bbf6b2aa28/entproto/adapter.go#L191
this is on the file level, i think you can do something similar on a field.

rotemtam:
> > @rotemtam ,
> > [AIP-132](https://google.aip.dev/132) states:
> > > The comment above the page_size field should document the maximum allowed value, as well as the default value if the field is omitted (or set to 0).
> > 
> > 
> > I was trying to determine how I could attach comments to the proto fields through the FieldDescriptorProto with no success. Do you know of a way to do this?
> 
> i think the pb lib we're using has this concept:
> 
> https://github.com/ent/contrib/blob/712b44424d50a097eea13d843148a7bbf6b2aa28/entproto/adapter.go#L191
> 
> 
> this is on the file level, i think you can do something similar on a field.

this isn't critical you can add it in a subsequent PR. 

also can you join [Ent slack](https://entgo.io/docs/slack)? I wanted to talk to you about a blog post about your great work here

JeremyV2014:
One other thing, do we want an option for selecting ascending/descending order in this implementation? Or should that be a separate PR?

JeremyV2014:
Actually, I forgot that users can set the type of the ID to something other than int and uuid. What is the exhaustive list of types that the ID can be? Is it just int, uuid, and string? It actually might be simpler to use gob for serializing/de-serializing a token struct if the list of types is much larger than that.

entproto/cmd/protoc-gen-entgrpc/template/method_list.tmpl
```diff
@@ -0,0 +1,79 @@
...
+            pageToken := int(token)
+        {{ else if .G.EntType.ID.IsUUID }}
+            pageToken, err := {{ qualify "github.com/google/uuid" "ParseBytes" }}(bytes)
+            if err != nil {
+                return nil, status.Errorf(codes.InvalidArgument, "page token is invalid")
+            }
+        {{ else if .G.EntType.ID.IsString }}
+            pageToken := string(bytes)
+        {{ else }}
+            {{ throwError "unsupported id type for list method" }}
```
a8m, line 38:
```suggestion
            {{ fail "unsupported id type for list method" }}
```


Use the builtin `fail` instead

entproto/cmd/protoc-gen-entgrpc/main.go
```diff
@@ -131,6 +131,9 @@ func (g *serviceGenerator) generate() error {
 					Method: m,
 				}
 			},
+			"throwError": func(description string) (string, error) {
+				return "", fmt.Errorf("entproto: %s", description)
+			},
```
a8m, line 136:
```suggestion
```

a8m: already exists in ent codegen functions.

entproto/cmd/protoc-gen-entgrpc/template/method_list.tmpl
```diff
@@ -0,0 +1,79 @@
...
+        {{ end }}
```
a8m, line 39:
```suggestion
        {{- if .G.EntType.ID.IsInt }}
            token, err := {{ qualify "strconv" "ParseInt" }}(string(bytes), 10, 32)
            if err != nil {
                return nil, status.Errorf(codes.InvalidArgument, "page token is invalid")
            }
            pageToken := int(token)
        {{- else if .G.EntType.ID.IsUUID }}
            pageToken, err := {{ qualify "github.com/google/uuid" "ParseBytes" }}(bytes)
            if err != nil {
                return nil, status.Errorf(codes.InvalidArgument, "page token is invalid")
            }
        {{- else if .G.EntType.ID.IsString }}
            pageToken := string(bytes)
        {{- else }}
            {{ throwError "unsupported id type for list method" }}
        {{- end }}
```

a8m: Let's remove redundant new lines between blocks.

entproto/cmd/protoc-gen-entgrpc/template/method_list.tmpl
```diff
@@ -0,0 +1,79 @@
...
+            {{ end }}
+            All(ctx)
+    }
+    switch {
+    case err == nil:
+        var nextPageToken string
+        if len(entList) == pageSize + 1 {
+			nextPageToken = {{ qualify "encoding/base64" "StdEncoding.EncodeToString" }}(
+			    []byte({{ qualify "fmt" "Sprintf" }}("%v", entList[len(entList)-1].ID)))
+			entList = entList[:len(entList)-1]
```
a8m, line 62:
```suggestion
		nextPageToken = {{ qualify "encoding/base64" "StdEncoding.EncodeToString" }}(
		    []byte({{ qualify "fmt" "Sprintf" }}("%v", entList[len(entList)-1].ID)))
		entList = entList[:len(entList)-1]
```

entproto/service.go
```diff
@@ -122,7 +127,8 @@ func (a *Adapter) genMethodProtos(genType *gen.Type, m Method) (methodResources,
 		Type:     &protoMessageFieldType,
 		TypeName: &genType.Name,
 	}
-	var output, methodName string
+	var outputName, methodName string
+	var messages []*descriptorpb.DescriptorProto
```
a8m, line 131:
```suggestion
	var (
		outputName, methodName string
		messages []*descriptorpb.DescriptorProto
	)
```

a8m (approved):
Looks good. minor comments

a8m:
Amazing contribution @JeremyV2014. Thanks a lot for your help, it's appreciated!

rotemtam (approved):
Great job, @JeremyV2014 !!

rotemtam:
@JeremyV2014 there's a merge conflict with your bugfix from today. Please fix it in this branch and we're good to go !

JeremyV2014:
@rotemtam all set!

rotemtam (approved):

PR #166, opened 2021-11-18
entoas: replace depth logic

masseelch:
can reintroduce max depth any time later if some user needs it.

cc @a8m


a8m (approved):

a8m:
Thanks @masseelch ❤️

PR #168, opened 2021-11-23
entoas: add annotation to support custom oas types

masseelch:
This PR adds an Annotation to create an OAS type to use for custom GoTypes.

### DIFF
--- a/entoas/annotation.go
+++ b/entoas/annotation.go
@@ -70,7 +70,7 @@ func OperationPolicy(p Policy) OperationConfigOption {
 // Example returns an example annotation.
 func Example(v interface{}) Annotation { return Annotation{Example: v} }
 
-// OASType returns a OASType annotation.
+// OASType returns an OASType annotation.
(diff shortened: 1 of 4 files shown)


entoas/annotation.go
```diff
@@ -67,6 +70,9 @@ func OperationPolicy(p Policy) OperationConfigOption {
 // Example returns an example annotation.
 func Example(v interface{}) Annotation { return Annotation{Example: v} }
 
+// OASType returns a OASType annotation.
```
a8m, line 73:
```suggestion
// OASType returns an OASType annotation.
```

a8m (approved):
Looks good. Don't you want to add a test case for this?

masseelch:
> Looks good. Don't you want to add a test case for this?

There is [one](https://github.com/ent/contrib/blob/a4942a59d7d8f7438ddb598fe559347f10f58df0/entoas/generator_test.go#L64). Or do you mean an integration one?

PR #177, opened 2021-12-04
Update node.tmpl to not include unreachable code

crossworth:
Hello, I'm facing a problem with `go vet` about this line:

https://github.com/ent/contrib/blob/0e3a610c60241fa56a27d1a423cea4c904f12dfd/entgql/template/node.tmpl#L293-L295

`pkg/ent/node.go:190:2: unreachable code`

image

As I understand it, it is not necessary as the `default` works as a catch all, doesn't?

### DIFF
--- a/entgql/internal/todo/ent/gql_node.go
+++ b/entgql/internal/todo/ent/gql_node.go
@@ -374,6 +374,7 @@ func (c *Client) noders(ctx context.Context, table string, ids []int) ([]Noder, 
 				*noder = node
 			}
 		}
+		return noders, nil
 	case todo.Table:
 		nodes, err := c.Todo.Query().
 			Where(todo.IDIn(ids...)).
@@ -387,10 +388,10 @@ func (c *Client) noders(ctx context.Context, table string, ids []int) ([]Noder, 
 				*noder = node
 			}
 		}
+		return noders, nil
 	default:
 		return nil, fmt.Errorf("cannot resolve noders from table %q: %w", table, errNodeInvalidID)
 	}
-	return noders, nil
 }
 
 type tables struct {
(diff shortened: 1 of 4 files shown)


alexsn (approved):

masseelch:
@crossworth Can you please run `go generate ./...` to regenerate all the assets changed by the template change?

a8m (commented):
Hey @crossworth, what version of contrib do you use?

I’m not sure there’s an issue with the current generated code. See https://github.com/ent/contrib/blob/master/entgql/internal/todo/ent/gql_node.go#L357

crossworth:
Hello, I use: `entgo.io/contrib v0.1.0`

Indeed the code looks fine.

But my generated code still fails, I don't have any `$gqlNodes` (I don't plan to use the `Noder` for now).

image

Maybe we should return from each case like the [`noder`](https://github.com/ent/contrib/blob/0e3a610c60241fa56a27d1a423cea4c904f12dfd/entgql/internal/todo/ent/gql_node.go#L290)?

masseelch:
One solution is to `return nodes` from every case. Then we won't end up with the case you have there.

alexsn:
Meh, thats ugly. I suggest that for the case where we have no node objects in schema we egenerate a one liner function always returning an error.

a8m:
> Meh, thats ugly. I suggest that for the case where we have no node objects in schema we generate a one liner function always returning an error.

Agree. Let's do it.

crossworth:
@masseelch Thats a great question! I started using without skipping the nodes, but since I have only a few nodes for now I thought that would make more sense to allow as needed and "skipped" all the nodes.

But now that I think about, this don't make sense, it's easier to disable the extension for now and enable as needed.

I will close the issue. Thank for the support guys.

masseelch:
I am glad your issue is resolved 🙂

I still think we should implement @alexsn proposed solution to make it "correct" even though the use case is not really given. The current implementation won't stand fuzzy testing (sometimes). Thoughts?

alexsn:
Another alternative would be to fail to render the node.tmpl if we have no nodes in schema.

masseelch:
I'll create a new PR tackling this.

PR #179, opened 2021-12-11
entoas: add primitive schemas

masseelch:
This is the first PR from multiple to use the spec files and builders provided by `ogen` instead of the ones provided by `entoas`.

The `ogen` spec is near complete regarding Specification compliance and thus will be `entoas`. The builders provided by `ogen` makes editing the generated Spec a breeze. 

This changes are BC. What are your intentions regarding this? 

@ernado If you want to have a look at this too and provide feedback, please do so :-)

Edit: The pipelines do not complete since I cannot make a small diff if fixing them. Before the merge will be completed into master they will be fixed ;-)


ernado:
This looks cool!

Will give a closer look tomorrow, but overall I like it.

masseelch:
> This looks cool!
> 
> Will give a closer look tomorrow, but overall I like it.

:bow:

PR #181, opened 2021-12-15
[BC]: entoas: use ogen spec definitions instead of custom ones

masseelch:
This PR replaces custom OAS struct definitions with the ones provided by [ogen](github.com/ogen-go/ogen). This is a BC.

- [x] Remove `entoas/internal/spec` package
- [x] Update README to include API BC with link to this PR
- [x] Wait for ogen DSL to merge this and update go.mod accordingly


masseelch:
@a8m and @rotemtam: The major changes are in `generator.go`. This PR is bigger, but since there are very few changes in domain logic it should be fast to review.

PR #204, opened 2022-01-14
fix(entgql/template): marshal cursor values that implement driver.Valuer

tarrencev:
fixes an issue with Cursors when the value is a GoType


alexsn:
Please add a test for the use case it's support to handle.

alexsn (changes requested):
Please add a test for the use case it's support to handle.

PR #205, opened 2022-01-15
fix: remove package name for the error

giautm:
Let it be consistent with other error


giautm:
I really don't want to expose the package name to users like:
```
entgo.io/contrib/entgql/internal/todouuid/ent: empty predicate CategoryWhereInput
```

a8m (approved):

PR #209, opened 2022-01-22
feat(entgql): support custom NodeID field

giautm:


a8m:
Thanks for the contribution @giautm. Would you mind elaborate on the use case? 🙂

giautm:
> Thanks for the contribution @giautm. Would you mind elaborate on the use case? 🙂

the use case it supports is when I migrate the old Ent-based project to support GQL NodeID (really exposed a Restful API). So, I need to keep the current ID column and add a new `global_id` to the schema. Then, expose it via this config. In the example, I really demo the use case:

```gql
type Document implements Node {
  id: ID! @goField(name: "GlobalID")
  legacyID: ID! @goField(name: "ID")
  name: String!
}
```

entgql/annotation.go
```diff
@@ -22,6 +22,8 @@ import (
 
 // Annotation annotates fields and edges with metadata for templates.
 type Annotation struct {
+	// NodeIDField is the name of the field that used as NodeID.
+	NodeIDField string
```
a8m, line 26: Let's keep it only `NodeID`.

giautm:
Hey @a8m, I renamed `NodeIDField` to `NodeID` and rebase the PR to the master branch. Will you consider merging it?

PR #217, opened 2022-02-03
entgql: enable bind by default

giautm:
`Bind()` become a no-op function to avoid breaking the existing schema.

This commit also added `Unbind()` to disable bind.

### DIFF
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -50,7 +50,10 @@ func OrderField(name string) Annotation {
 // No-op function to avoid breaking the existing schema.
 // You can safely remove this function from your scheme.
 // To disable Bind, use BindDisabled()
-func Bind() Annotation {
+func Bind(binds ...bool) Annotation {
+	if len(binds) > 0 {
+		return Annotation{BindDisabled: !binds[0]}
+	}
 	return Annotation{}
 }
 
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -24,9 +24,9 @@ import (
 type Annotation struct {
 	// OrderField is the ordering field as defined in graphql schema.
 	OrderField string `json:"OrderField,omitempty"`
-	// Bind implies the edge field name in graphql schema
+	// BindDisabled implies the edge field name in graphql schema
 	// is equivalent to the name used in ent schema.
-	Bind bool `json:"Bind,omitempty"`
+	BindDisabled bool `json:"BindDisabled,omitempty"`
 	// Mapping is the edge field names as defined in graphql schema.
 	Mapping []string `json:"Mapping,omitempty"`
 	// Type is the underlying GraphQL type name (e.g. Boolean).
@@ -46,8 +46,17 @@ func OrderField(name string) Annotation {
 }
 
 // Bind returns a binding annotation.
+//
+// No-op function to avoid breaking the existing schema.
+// You can safely remove this function from your scheme.
+// To disable Bind, use BindDisabled()
 func Bind() Annotation {
-	return Annotation{Bind: true}
+	return Annotation{}
+}
+
+// BindDisabled returns a disabled binding annotation.
+func BindDisabled() Annotation {
+	return Annotation{BindDisabled: true}
 }
 
 // MapsTo returns a mapping annotation.
@@ -81,8 +90,8 @@ func (a Annotation) Merge(other schema.Annotation) schema.Annotation {
 	if ant.OrderField != "" {
 		a.OrderField = ant.OrderField
 	}
-	if ant.Bind {
-		a.Bind = true
+	if ant.BindDisabled {
+		a.BindDisabled = true
 	}
 	if len(ant.Mapping) != 0 {
 		a.Mapping = ant.Mapping
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -24,8 +24,10 @@ import (
 type Annotation struct {
 	// OrderField is the ordering field as defined in graphql schema.
 	OrderField string `json:"OrderField,omitempty"`
-	// Unbind implies the edge field name in graphql schema
-	// isn't equivalent to the name used in ent schema.
+	// Unbind implies the edge field name in GraphQL schema is not equivalent
+	// to the name used in ent schema. That means, by default, edges with this
+	// annotation will not be eager-loaded on Paginate calls. See the `MapsTo`
+	// option in order to load edges be different name mapping.
 	Unbind bool `json:"Unbind,omitempty"`
 	// Mapping is the edge field names as defined in graphql schema.
 	Mapping []string `json:"Mapping,omitempty"`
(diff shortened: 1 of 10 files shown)


entgql/internal/todo/ent/schema/todo.go
```diff
@@ -69,7 +69,9 @@ func (Todo) Edges() []ent.Edge {
 			Annotations(entgql.Bind()).
 			Unique(),
 		edge.From("category", Category.Type).
+			Annotations(entgql.BindDisabled()).
 			Ref("todos").
+			Annotations(entgql.BindDisabled()).
```
giautm, line 74: This is a demonstration for using `BindDisabled()`, it keeps the generated code is the same as before.

giautm:
This commit still is a breaking change for who using Mapping() anonations, they required to use BindDisabled() to fix the break.

giautm:
I think we should merge Mapping and Bind to a single field, with two values: binding/mapping.

entgql/annotation.go
```diff
@@ -46,8 +46,17 @@ func OrderField(name string) Annotation {
...
+// To disable Bind, use BindDisabled()
 func Bind() Annotation {
-	return Annotation{Bind: true}
+	return Annotation{}
+}
+
+// BindDisabled returns a disabled binding annotation.
+func BindDisabled() Annotation {
+	return Annotation{BindDisabled: true}
 }
```
a8m, line 61: What do you think about accepting an optional parameter as follows:

```go
edge.To(...).
    Annotation(
       entgql.Bind() // Default to true. 
    )

edge.To(...).
    Annotation(
       entgql.Bind(true) // Same as Bind().
    )

edge.To(...).
    Annotation(
       entgql.Bind(false) // Allow overriding the default behaviour.
    )
```

Then, you can keep the `Bind` field but as a pointer, or use BindDisabled.

giautm: Is Go has optional argument? Or you trying to said:

funct Bind(bind ...bool) {}

a8m: exactly.

giautm: > func Bind(bind ...bool) {}

I think it's a bit complicated to implement, and it's also kind of weird to use only the first element.

giautm: I'm updated the PR with this commit https://github.com/ent/contrib/pull/217/commits/3c7a9d28cc1b43f48d94ed293fb9a02f0b970ee5
But It can be lead to the invalid case when User using it as `Annotation(entgql.MapTo("field1", "field2"), entgql.Bind(true))`

entgql/annotation.go
```diff
@@ -46,13 +46,29 @@ func OrderField(name string) Annotation {
...
+	if len(binds) > 0 && !binds[0] {
+		return Annotation{BindDisabled: true}
+	}
+	return Annotation{}
+}
+
+// BindDisabled returns a disabled binding annotation.
+func BindDisabled() Annotation {
+	return Annotation{BindDisabled: true}
 }
```
a8m, line 63: Don't want to be annoying, but what do you think about `Unbind()` instead of `BindDisabled`? I'm very picky about the public API. If we add `Unbind` I think the change in `Bind` is not needed.

giautm: I'm okay with that name, update late. ;)

giautm: But, Bind() still No-op function, right?

a8m: Right.

entgql/annotation.go
```diff
@@ -24,9 +24,9 @@ import (
 type Annotation struct {
 	// OrderField is the ordering field as defined in graphql schema.
 	OrderField string `json:"OrderField,omitempty"`
-	// Bind implies the edge field name in graphql schema
-	// is equivalent to the name used in ent schema.
-	Bind bool `json:"Bind,omitempty"`
+	// Unbind implies the edge field name in graphql schema
+	// isn't equivalent to the name used in ent schema.
+	Unbind bool `json:"Unbind,omitempty"`
```
a8m, line 29:
```suggestion
	// Unbind implies the edge field name in GraphQL schema is not equivalent
	// to the name used in ent schema. That means, by default, edges with this
	// annotation will not be eager-loaded on Paginate calls. See the `MapsTo`
	// option in order to load edges be different name mapping.
	Unbind bool `json:"Unbind,omitempty"`
```

entgql/annotation.go
```diff
@@ -46,13 +46,25 @@ func OrderField(name string) Annotation {
 }
 
 // Bind returns a binding annotation.
+//
+// No-op function to avoid breaking the existing schema.
+// You can safely remove this function from your scheme.
+// To disable Bind, use Unbind()
```
a8m, line 52:
```suggestion
//
// Deprecated: the Bind option predates the Unbind option, and it is planned
// to be removed in future versions. Users should not use this annotation as it
// is a no-op call, or use `Unbind` in order to disable `Bind`.
```

entgql/annotation.go
```diff
@@ -46,13 +46,25 @@ func OrderField(name string) Annotation {
...
 func Bind() Annotation {
-	return Annotation{Bind: true}
+	return Annotation{}
+}
+
+// Unbind returns a unbinding annotation.
```
a8m, line 57:
```suggestion
// Unbind implies the edge field name in GraphQL schema is not equivalent
// to the name used in ent schema. That means, by default, edges with this
// annotation will not be eager-loaded on Paginate calls. See the `MapsTo`
// option in order to load edges be different name mapping.
```

entgql/annotation.go
```diff
@@ -46,13 +46,25 @@ func OrderField(name string) Annotation {
...
+func Unbind() Annotation {
+	return Annotation{Unbind: true}
 }
 
 // MapsTo returns a mapping annotation.
 func MapsTo(names ...string) Annotation {
-	return Annotation{Mapping: names}
+	return Annotation{
+		Mapping: names,
+		Unbind:  true, // Unbind because it cant be used with mapping names
```
a8m, line 66:
```suggestion
		Unbind:  true, // Unbind because it cant be used with mapping names.
```

entgql/template.go
```diff
@@ -78,6 +80,44 @@ func parseT(path string) *gen.Template {
...
+}
+
+func fieldCollections(edges []*gen.Edge) (map[string]fieldCollection, error) {
+	result := make(map[string]fieldCollection)
+	for _, e := range edges {
+		result[e.Name] = fieldCollection{
+			Name:    e.Type.Name,
+			Mapping: []string{e.Name},
+		}
+
```
a8m, line 95:
```suggestion
```

entgql/template.go
```diff
@@ -78,6 +80,44 @@ func parseT(path string) *gen.Template {
...
+				}
+
+				result[e.Name] = fieldCollection{
+					Name:    e.Type.Name,
+					Mapping: ant.Mapping,
+				}
+			}
+		}
+	}
+
```
a8m, line 117:
```suggestion
		if e.Annotations == nil || e.Annotations[ant.Name()] == nil {
			continue
		}
		if err := ant.Decode(e.Annotations[ant.Name()]); err != nil {
			return nil, err
		}
		if ant.Unbind {
			delete(result, e.Name)
		}
		if len(ant.Mapping) > 0 {
			if _, bind := result[e.Name]; bind {
				return nil, errors.New("bind and mapping annotations are mutually exclusive")
			}
			result[e.Name] = fieldCollection{
				Name:    e.Type.Name,
				Mapping: ant.Mapping,
			}
		}
	}
```

Love this! 😍

a8m (approved):
Love this! Thanks for adding this. 

Added a few minor comments (doc and style). Let's fix them and update ent/ent with the docs change. 🎉

a8m (approved):

a8m:
Thanks, @giautm! ❤️

PR #218, opened 2022-02-05
chore: added devcontainer

giautm:
This PR setup devcontainer for everyone who works on VSCode, it also installs all tools to run `go generate ./...` without any additional steps.


giautm:
Close, because only me using it

PR #219, opened 2022-02-05
Added support to generate byte fields.

anivanovic:
I was using entimport tool and was not able to generate schema files. Generating byte field was not supported for some reason. My column is defined as `BYTEA` in Postgres. After adding byte type to case switch entimport generated schema without problems


rotemtam (approved):
Thanks for your contribution @anivanovic !

cc @zeevmoney

PR #220, opened 2022-02-06
entgql: experimental global-id support with relay-format (<type>:<id>)

a8m:


giautm:
@theprobugmaker running to the case he needs to encode UUID as Global ID with the Entity Number: `<EntityNumber>\n<UUID>`.

I think there will be many different formats used, but it is consistent in the project. EntGQL should provide a template that allows to format it.

The user can write customs UnmarshalGQL/MarshalGQL functions to parse Type and ID from/to String.

theprobugmaker:
> @theprobugmaker running to the case he needs to encode UUID as Global ID with the Entity Number: `<EntityNumber>\n<UUID>`.
> 
> I think there will be many different formats used, but it is consistent in the project. EntGQL should provide a template that allows to format it.
> 
> The user can write customs UnmarshalGQL/MarshalGQL functions to parse Type and ID from/to String.

That is right, thank you.

PR #225, opened 2022-02-09
Add support for UUID fields.

anivanovic:
Adding support for generating UUID fields.
cc @a8m


rotemtam (approved):
💯

Thanks for your contribution!

PR #227, opened 2022-02-10
Add support for generating JSON fields

anivanovic:
Easiest implementation to support generating JSON fields I could think of.
Developers still need to change schema but at least generation does not fail
and generated file compiles.

### DIFF
--- a/schemast/printer_test.go
+++ b/schemast/printer_test.go
@@ -104,6 +104,19 @@ func TestPrintAddImport(t *testing.T) {
 	require.Equal(t, matches, 1)
 }
 
+func TestPrintStructOnSameLine(t *testing.T) {
+	tt, err := newPrintTest(t)
+	require.NoError(t, err)
+
+	require.NoError(t, tt.ctx.AppendField("Message", field.String("name").Descriptor()))
+	require.NoError(t, tt.ctx.AppendField("Message", field.JSON("json", struct{}{}).Descriptor()))
+	require.NoError(t, tt.print())
+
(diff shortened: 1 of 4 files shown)


anivanovic:
> @anivanovic I made a similar contribution, I addressed more cases with JSON field, would you like to talk in Discord in order to push it together?

Sure! Can you give me Discord link? I am also in Slack #ent channel. So we can open a thread there also.

a8m:
Thanks for the support @zeevmoney, and thanks @anivanovic for the contribution.
Here's a link to the Discord server. Feel free to create a thread there - https://discord.gg/qZmPgTE6RX

I also suggest you use `go/parser.ParseExpr` to construct and validate the expression for configuring the JSON type. Please, feel free to mention me in Discord 😃

schemast/printer_test.go
```diff
@@ -104,6 +104,19 @@ func TestPrintAddImport(t *testing.T) {
 	require.Equal(t, matches, 1)
 }
 
+func TestPrintStructOnSameLine(t *testing.T) {
```
anivanovic, line 107: Using `parser.ParseExp` breaks this check. Generated struct is put in file as
```
struct{
}{}
```

Do not know if this is non issue and we can live with or we would like to have formatting like before. All fields on one line.

zeevmoney: image

zeevmoney (approved):
great job!

a8m (approved):

a8m:
Thanks, @anivanovic for your contribution! ❤️

PR #229, opened 2022-02-10
fix: type ID check should run on non-skipped nodes

giautm:


giautm:
There is more issue with the current `node.tmpl`, `tables` field was dependent on global `$.IDType`, not updated after filterNodes so it generate wrong code (`/workspace/ent/client.go:31:9: undeclared name: tables`)

I updated this PR to fix the issue

https://github.com/ent/contrib/blob/2b75bf138815f589ede1dc72cc83ce3aa9206ebc/entgql/template/node.tmpl#L351-L357

a8m (approved):
💯

PR #231, opened 2022-02-11
entgql: generate create/update mutation inputs

giautm:
Related https://github.com/ent/ent/issues/2351


entgql/internal/todo/ent/gql_mutation_input.go
```diff
@@ -39,9 +39,7 @@ type CreateCategoryInput struct {
 func (i *CreateCategoryInput) Mutate(m *CategoryCreate) {
 	m.SetText(i.Text)
 	m.SetStatus(i.Status)
-	if v := i.Config; v != nil {
-		m.SetConfig(*v)
-	}
+	m.SetConfig(i.Config)
```
giautm: Updated to correct code

crossworth:
Hello @giautm , there a few things that I think we should change on the create mutations template that I have been fixing on my local copy and would like to let you know.

- The skip annotation should not be used to skip the mutation generation, a new annotation should be created only for the mutation part, like, I like to be able to order/filter by `created_time`, `updated_time`, but I dont want to allow a mutation on it.
- How to solve the problem of a null field vs a a field not provided, is the `ClearType bool` the best solution? Maybe we should try read the provided query and handle in a different way? (to make the javascript part more easy/compatible).

giautm:
> I like to be able to order/filter by created_time, updated_time, but I dont want to allow a mutation on it.

You can use `entgql.Skip()` on each fields you want to ignore.

giautm:
> * I like to be able to order/filter by `created_time`, `updated_time`, but I dont want to allow a mutation

Hey @crossworth, I think the template really depends on `MutableFields`.

giautm:
hello @crossworth now, you can use `entgql.Skip(entgql.SkipMutationCreateInput)` to skip generate Create<Type>Input!

entgql/schema.go
```diff
@@ -478,6 +488,128 @@ func (e *schemaGenerator) buildWhereInput(t *gen.Type, gqlType string) (*ast.Def
...
+			if ant.Skip.Is(SkipMutationCreateInput) {
+				continue
+			}
+
+			if e.Unique {
+				def.Fields = append(def.Fields, &ast.FieldDefinition{
+					Name: camel(e.Name + "_ID"),
+					Type: namedType("ID", true),
+				}, &ast.FieldDefinition{
+					Name: camel(snake(e.MutationClear())),
```
giautm, line 593: @a8m: I'm not sure it outputs the same result.

entgql/schema.go
```diff
@@ -478,6 +488,128 @@ func (e *schemaGenerator) buildWhereInput(t *gen.Type, gqlType string) (*ast.Def
...
+					Type: namedType("Boolean", true),
+				})
+			} else {
+				def.Fields = append(def.Fields, &ast.FieldDefinition{
+					Name: camel("add_" + singular(e.Name) + "_IDs"),
+					Type: namedType("[ID!]", true),
+				}, &ast.FieldDefinition{
+					Name: camel("remove_" + singular(e.Name) + "_IDs"),
+					Type: namedType("[ID!]", true),
+				})
```
giautm, line 603: Due to a "bug" with "IDs", I can't use `MutationAdd()` and `MutationRemove()` functions

`camel(snake(e.MutationRemove()))` return `removeGroupIds`
and `camel(e.MutationRemove())` return `removegroupids`

So, I have to change copy it logic: `camel("remove_" + singular(e.Name) + "_IDs")`

entgql/schema.go
```diff
@@ -229,6 +231,15 @@ func (e *schemaGenerator) buildTypes(g *gen.Graph, s *ast.Schema) error {
...
 			}
 		}
+
+		if e.genMutations {
+			defs, err := e.buildMutationInputs(node, ant, gqlType)
+			if err != nil {
+				return err
+			} else if len(defs) > 0 {
+				s.AddTypes(defs...)
+			}
```
a8m, line 241:
```suggestion
			defs, err := e.buildMutationInputs(node, ant, gqlType)
			if err != nil {
				return err
			}
			if if len(defs) > 0 {
				s.AddTypes(defs...)
			}
```
No need for else.

entgql/schema.go
```diff
@@ -478,6 +489,129 @@ func (e *schemaGenerator) buildWhereInput(t *gen.Type, gqlType string) (*ast.Def
...
+			def.Fields = append(def.Fields, &ast.FieldDefinition{
+				Name:        camel(f.Name),
+				Type:        namedType(scalar, true),
+				Description: f.Comment(),
+			})
+			if f.Optional {
+				def.Fields = append(def.Fields, &ast.FieldDefinition{
+					Name: camel("clear_" + f.Name),
+					Type: namedType("Boolean", true),
+				})
```
a8m, line 578:
```suggestion
				def.Fields = append(def.Fields, &ast.FieldDefinition{
					Name: "clear" + f.StructField(),
					Type: namedType("Boolean", true),
				})
```

entgql/schema.go
a8m: "clear" + f.StructField()

entgql/schema.go
```diff
@@ -478,6 +489,129 @@ func (e *schemaGenerator) buildWhereInput(t *gen.Type, gqlType string) (*ast.Def
...
+			if err != nil {
+				return nil, err
+			}
+			if ant.Skip.Is(SkipMutationCreateInput) {
+				continue
+			}
+
+			if e.Unique {
+				def.Fields = append(def.Fields, &ast.FieldDefinition{
+					Name: camel(e.Name + "_ID"),
```
a8m, line 592: ditto

giautm, line 592:
```suggestion
					Name: camel(e.Name) + "ID",
```

entgql/schema.go
giautm: Screen Shot 2022-05-12 at 22 13 31
I think also need to change the go template.

entgql/schema.go
```diff
@@ -478,6 +490,100 @@ func (e *schemaGenerator) buildWhereInput(t *gen.Type, gqlType string) (*ast.Def
...
+						Type: namedType("[ID!]", e.Optional),
+					})
+				} else {
+					def.Fields = append(def.Fields, &ast.FieldDefinition{
+						Name: camel("add_" + singular(e.Name) + "_IDs"),
+						Type: namedType("[ID!]", true),
+					}, &ast.FieldDefinition{
+						Name: camel("remove_" + singular(e.Name) + "_IDs"),
+						Type: namedType("[ID!]", true),
+					})
```
a8m, line 577:
```suggestion
						Name: "add" + pascal(singular(e.Name)) + "IDs",
						Type: namedType("[ID!]", true),
					}, &ast.FieldDefinition{
						Name: "remove" + pascal(singular(e.Name)) + "IDs",
						Type: namedType("[ID!]", true),
					})
```

entgql/annotation.go
```diff
@@ -335,6 +350,35 @@ func (a queryFieldAnnotation) Directives(directives ...Directive) queryFieldAnno
 	return a
 }
 
+type MutationOption interface {
+	IsCreate() bool
+}
+
+type builtinMutation bool
+
+func (v builtinMutation) IsCreate() bool { return bool(v) }
```
a8m, line 359: Why not use functional options that can change the `Options` object (instead of interface). Feels easier to write.

a8m (approved):

PR #234, opened 2022-02-15
entgql: remove unnecessary `Funcs` call

giautm:
Because in `gen.NewTemplate()` we already call `t.Funcs(Funcs)`

https://github.com/ent/ent/blob/a7beb5c834f0b81de7afd15350da9c42d8eb6b88/entc/gen/template.go#L231-L235


a8m (approved):
💯 

Thanks @giautm ❤️

PR #235, opened 2022-02-16
entgql: create plugin for GQLGen

giautm:
Improve ent + gqlgen interop

Features:

* [x] Automatically map objects
* [x] Generate Relay objects
* [x] Custom scalars
* [x] Map Pagination (Order by) Input Objects
* [x] Map Enums (this PR)
* [x] Allow to set GQL description
* [x] Auto config ID type
* [ ] Custom Cursor type
* [ ] Map Edges
* [ ] Rename fields
* [ ] ~Support Default Values~  Only makes sense for inputs

`EntGQL` extension changes:
* [ ] the templates respect `entgql.RelayConnection()` for generate `<Name>Edge`, `<Name>Connection`, `<Name>Order`, and `<Name>OrderField`.
* [ ] Move `WithWhereFilters` feature to EntGQL plugin, because it should generate both Go codes and GQL types. (?!)


giautm:
Finally, It's green!!!!

cliedeman:
@giautm glad to see you working on this. I tried a few days ago to update the plugin and failed 😀.

Shout if you have any questions.

I think you are also a relay user. I also used the relay Type:ID encoded style and had another plugin that made gqlgen use a RelayID resolve to return the encoded ID too. That may be useful as a switch but this is a great starting point

giautm:
> I also used the relay Type:ID encoded style and had another plugin that made gqlgen use a RelayID resolve to return the encoded ID too

Hey, can you share that plugin? I and @crossworth are working on it. If you can give some examples that will be great. 😄

giautm:
@cliedeman `Custom scalars` => This feature looks like don't work. Because I can't pass `GqlScalarMappings` to the graph

cliedeman:
Sure. 

Here are the relevant pieces

relay.tmpl
```
{{/* gotype: entgo.io/ent/entc/gen.Graph */}}
{{ define "stackworx_relay" }}

{{ $pkg := base $.Config.Package }}

{{ template "header" $ }}

import (
	"context"
	"encoding/base64"
	"fmt"
	"strconv"
	"strings"

	{{- range $n := $.Nodes }}
		"{{ $.Config.Package }}/{{ $n.Package }}"
	{{- end }}
)

func toRelayID(t string, id int) string {
	return base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%d", t, id)))
}

{{ range $n := $.Nodes -}}
{{ $r := $n.Receiver }}
    // RelayID encodes the objects id into a relay globally unique id
    func ({{ $r }} *{{ $n.Name }}) RelayID(ctx context.Context) (string, error) {
        return toRelayID("{{ $n.Name }}", {{ $r }}.{{ pascal $n.ID.Name }}), nil
    }
{{ end }}


func RelayTypeToTable(t string) (string, error) {
    switch t {
    {{- range $n := $.Nodes }}
    case "{{ $n.Name }}":
        return {{ $n.Package }}.Table, nil
    {{- end }}
    default:
        return "", fmt.Errorf("unknown type: %s", t)
    }
}

func FromRelayID(id string) (string, string, error) {
	raw, err := base64.StdEncoding.DecodeString(id)
	if err != nil {
		return "", "", fmt.Errorf("relayId: failed to decode: %w", err)
	}
	s := strings.Split(string(raw), ":")
	if len(s) != 2 {
		return "", "", fmt.Errorf("relayId: invalid format: %s", raw)
	}
	return s[0], s[1], nil
}

{{ end }}
```

(this might be superseded by the PR @a8m is working on

relayid.go
```
package gqlgencontrib

import (
	_ "embed"
	"entgo.io/ent/entc/gen"
	"github.com/99designs/gqlgen/codegen"
	"github.com/99designs/gqlgen/plugin"
)

type relayIdPlugin struct {
	graph *gen.Graph
}

func NewRelayIdPlugin(graph *gen.Graph) plugin.Plugin {
	return relayIdPlugin{
		graph: graph,
	}
}

func implementsNodeInterface(obj *codegen.Object) bool {
	for _, i := range obj.Interfaces {
		if i == "Node" {
			return true
		}
	}
	return false
}

func (r relayIdPlugin) GenerateCode(data *codegen.Data) error {
	for _, o := range data.Objects {
		if implementsNodeInterface(o) {
			for _, f := range o.Fields {
				if isIdField(f) && r.isEntNode(o.Name) {
					f.GoFieldName = "RelayID"
					f.IsResolver = false
					f.GoFieldType = codegen.GoFieldMethod
					f.MethodHasContext = true
				}
			}
		}
	}
	return nil
}

func isIdField(f *codegen.Field) bool {
	return f.Name == "id" &&
		f.Type.Name() == "ID" &&
		f.Type.NonNull
}

func (r relayIdPlugin) Name() string {
	return "Stackworx:relayId"
}

func (r relayIdPlugin) isEntNode(name string) bool {
	// TODO: confirm node is not skipped - https://github.com/ent/contrib/pull/76
	for _, n := range r.graph.Nodes {
		if n.Name == name {
			return true
		}
	}
	return false
}

var _ plugin.Plugin = &relayIdPlugin{}
var _ plugin.CodeGenerator = &relayIdPlugin{}
```

This is the key part. When the above conditions are met we override the resolver function
```go
f.GoFieldName = "RelayID"
					f.IsResolver = false
					f.GoFieldType = codegen.GoFieldMethod
					f.MethodHasContext = true
```

cliedeman:
> @cliedeman Custom scalars => This feature looks like don't work. Because I can't pass GqlScalarMappings to the graph

What exactly is going wrong? I think the whole ideal with that annotation was to bulk map types to some graphql scalar. E.g. auto map all URL's

giautm:
> > @cliedeman Custom scalars => This feature looks like don't work. Because I can't pass GqlScalarMappings to the graph
> 
> What exactly is going wrong? I think the whole ideal with that annotation was to bulk map types to some graphql scalar. E.g. auto map all URL's

https://github.com/ent/contrib/pull/235/commits/52ec0b62f93ebd97fece07b7e158c9e233760bd5#diff-a965369faa5de2306c58d53e36c7393f64a0bff88380dba01c986d7b2330116cR40-R51

I found how to use `GQLScalarMappings` with the plugin. But I think it's complicated. If we don't use it in the EntGQL extension template, then we can pass it as a plugin option for simplicity.

cliedeman:
its very possible it got superseded by some by some other work. 😅 I wrote that code so long ago I am struggling to remember my plan

giautm:
@a8m: A question about API design for `Rename fields`. There is an annotation called `entgql.MapsTo` for mapping Edge names. So, Can we reuse it for the fields? or create a new annotation `GQLName`?

entgql/annotation.go
```diff
@@ -35,6 +35,16 @@ type Annotation struct {
 	Type string `json:"Type,omitempty"`
 	// Skip exclude the type
 	Skip bool `json:"Skip,omitempty"`
+	// Description is description of the GQL field/type.
+	Description string `json:"Description,omitempty"`
+	// Directives to add on the field/type.
+	Directives []Directive `json:"Directives,omitempty"`
+	// RelayConnection expose this node as a relay connection.
+	RelayConnection bool `json:"RelayConnection,omitempty"`
```
cliedeman, line 43: it might make sense to expand this into a more full config object.

E.g.

```go
type RelayConnectionConfig struct {
  Enabled bool
  IncludeTotal bool
  EdgeNotNull bool
  NodeNotNull bool
}
```

entgql/plugin/relay.go
```diff
@@ -0,0 +1,121 @@
...
+				Type:        ast.ListType(ast.NamedType(fmt.Sprintf("%sEdge", t.Name), nil), nil),
+				Description: "A list of edges.",
+			},
+			{
+				Name:        "pageInfo",
+				Type:        ast.NonNullNamedType("PageInfo", nil),
+				Description: "Information to aid in pagination.",
+			},
+			{
+				Name: "totalCount",
```
cliedeman, line 116: nit: this should be configurable

entgql/annotation.go
giautm: `Edge` and `Node` nullability was defined in the relay spec. I don't think we need options for them.

entgql/plugin/relay.go
giautm: Is there any reason that we should not expose this field? It's not defined in the relay spec but it's exists in `graphql-relay-js`

entgql/plugin/relay.go
```diff
@@ -0,0 +1,121 @@
...
+		Description: "An edge in a connection.",
+		Fields: []*ast.FieldDefinition{
+			{
+				Name:        "node",
+				Type:        ast.NamedType(t.Name, nil),
+				Description: "The item at the end of the edge",
+			},
+			{
+				Name:        "cursor",
+				Type:        ast.NonNullNamedType("Cursor", nil),
```
cliedeman, line 95: nit: it is not mandatory for the cursor to be nullable- https://relay.dev/graphql/connections.htm#sec-Cursor.

When I can I will actually make my connections as strict as possible

E.g.

```graphql
TodoConnection {
  edges [{
    node: Todo!
    cursor!
  }]!
  ...
}

but this is not mandatory

**Review by cliedeman — APPROVED** [review]
@giautm Looks good to me. I marked the comments as nit becaues the PR is big enough as it is and that can be handled in a follow up

File: entgql/annotation.go (hunk 1 above)
Comment (cliedeman) [reply entgql/annotation.go:43]: but there is no way for the user to control them - but this might be ok.

File: entgql/plugin/relay.go (hunk 2 above)
Comment (cliedeman) [reply entgql/plugin/relay.go:116]: just to give the user the option to hide it if they wanted too

File: entgql/plugin/relay.go (hunk 3 above)
Comment (giautm) [reply entgql/plugin/relay.go:95]: I marked the cursor is non-null type, so I dont understand what the problem is?

**Comment by giautm** [comment]
I have a consider about how the plugin and Ent can be interop, the plugin depended on ent.Graph to build the GQL scheme; Currently, it's created via `entc.LoadGraph`, but it looks disconnected with Ent's generate.  Can you improve it by changing some Ent/GQLGen API? To understand my point, please look into my `main.go` which attached bellow:

```go
const gqlgenConfigFile = "./gqlgen.yml"

func main() {
	graph, err := entc.LoadGraph("./ent/schema", &gen.Config{})
	if err != nil {
		log.Fatalf("entc: failed loading ent graph: %v", err)
	}

	entGQLPlugin, err := plugin.New(graph)
	if err != nil {
		log.Fatalf("entc: creating EntGQL plugin: %v", err)
	}

	gqlgenOpts := []api.Option{
		api.PrependPlugin(entGQLPlugin),
	}
	exEntGQL, err := entgql.NewExtension(
		entgql.WithWhereFilters(true),
		entgql.WithConfigPath(gqlgenConfigFile, gqlgenOpts...),
		// Generate the filters to a separate schema
		// file and load it in the gqlgen.yml config.
		entgql.WithSchemaPath("./graphql/schema/ent.gql"),
	)
	if err != nil {
		log.Fatalf("entc: creating EntGQL extension: %v", err)
	}

	exFederation, err := gqlfederation.NewExtension()
	if err != nil {
		log.Fatalf("entc: creating GQLFederation extension: %v", err)
	}
	err = entc.Generate("./ent/schema", &gen.Config{},
		entc.Extensions(exEntGQL, exFederation),
		entc.TemplateDir("./ent/template"),
	)
	if err != nil {
		log.Fatalf("entc: running ent codegen: %v", err)
	}

	cfg, err := config.LoadConfig(gqlgenConfigFile)
	if err != nil {
		log.Fatalf("gqlgen: failed to load config: %v", err)
	}

	if err = api.Generate(cfg, gqlgenOpts...); err != nil {
		log.Fatalf("gqlgen: running generate: %v", err)
	}
}
```
Source: https://github.com/giautm/workspace-entgo/blob/99364d8216d4d5c5e10975504762130d84f21403/cmd/tools/codegen/main.go

PR #236, opened 2022-02-16
entgql: update gqlgen version to latest master

giautm:


tools.go
```diff
@@ -0,0 +1,5 @@
+package contrib
+
+import (
+	_ "github.com/google/addlicense"
```
giautm, line 4: I added this file to keep `github.com/google/addlicense` in the `go.mod`. Without it, `go.mod` and `go.sum` always change after run `go mod tidy` which is the default behavior of GQLGen.

a8m (approved):

PR #247, opened 2022-03-06
entgql: use goModel directive for mapping

a8m:
In case the GraphQL node name is different from the Ent name, `autobind` won't work, and we need to map it explicitly.


entgql/schema.go
```diff
@@ -24,12 +24,60 @@ import (
...
+	nodes     []*gen.Type
+	relaySpec bool
+}
+
+var (
+	pos        = &ast.Position{Src: &ast.Source{BuiltIn: false}}
+	directives = map[string]*ast.DirectiveDefinition{
+		"goModel": {
+			Name:     "goModel",
+			Position: pos,
```
giautm, line 39: position can be set to nil

a8m: It's needed here, but not in other places. 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x150c2c2]

goroutine 1 [running]:
github.com/vektah/gqlparser/v2/formatter.(*formatter).FormatDirectiveDefinition(0x15a7f60, 0xc000347a40)
```

https://github.com/vektah/gqlparser/blob/master/formatter/formatter.go#L326-L328

giautm (approved):
LGTM

giautm:
How to mapping the custom type CategoryConfig to GoType with the goModel directive?

a8m:
> How to mapping the custom type CategoryConfig to GoType with the goModel directive?

I'll map fields with custom GoType in a different PR.

PR #250, opened 2022-03-06
entgql: support builtin JSON types

giautm:
Included: Ints(), Floats(), Strings()

### DIFF
--- a/entgql/schema.go
+++ b/entgql/schema.go
@@ -16,6 +16,7 @@ package entgql
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"entgo.io/ent/entc/gen"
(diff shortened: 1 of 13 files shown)


entgql/schema.go
```diff
@@ -324,6 +332,19 @@ func (e *schemaGenerator) typeFromField(f *gen.Field, idField bool, userDefinedT
...
+		if f.Type.RType != nil {
+			switch f.Type.RType.Kind {
+			case reflect.Slice, reflect.Array:
+				switch f.Type.RType.Ident {
+				case "[]float64":
+					return listNamedType("Float", f.Optional), nil
+				case "[]int":
+					return listNamedType("Int", f.Optional), nil
+				case "[]string":
+					return listNamedType("String", f.Optional), nil
```
a8m, line 344: This is generated as `[String!]`, but the `Optional` attribute is on the slice, not its elements. i.e. `[String]!`

giautm: We can't put a nil value in the string slide, right?

a8m (approved):

PR #249, opened 2022-03-06
entgql: add goModel directive to generated enums

a8m:
Next PR, will add support for mapping fields with custom GoType to `goModel`.

PR #253, opened 2022-03-07
entgql: generate custom scalar

giautm:
Related: https://github.com/ent/ent/issues/2351


entgql/schema.go
```diff
@@ -78,26 +79,72 @@ var (
...
+	return &ast.Directive{
+		Name: "goModel",
+		Arguments: ast.ArgumentList{
+			{
+				Name:  "model",
+				Value: &ast.Value{Kind: ast.StringValue, Raw: model},
+			},
+		},
+	}
+}
```
giautm, line 120: This may conflict with PR: https://github.com/ent/contrib/pull/249

entgql/extension.go
```diff
@@ -42,6 +42,14 @@ import (
 )
 
 type (
+	// CustomScalar holds information about a custom scalar type.
+	CustomScalar struct {
+		Name           string
+		Description    string
+		SpecifiedByURL string
+		GoModel        string
```
giautm, line 50: Ident or GoModel?

entgql/extension.go
```diff
@@ -201,6 +209,14 @@ func WithSchemaGenerator() ExtensionOption {
 	}
 }
 
+// WithCustomScalars adds custom scalar types to the schema.
+func WithCustomScalars(scalars ...CustomScalar) ExtensionOption {
+	return func(e *Extension) error {
+		e.scalars = scalars
```
giautm, line 215: Todo: add validate

entgql/extension.go
```diff
@@ -201,6 +209,14 @@ func WithSchemaGenerator() ExtensionOption {
 	}
 }
 
+// WithCustomScalars adds custom scalar types to the schema.
+func WithCustomScalars(scalars ...CustomScalar) ExtensionOption {
```
a8m, line 213: Why do you need this? This information can be extracted from the GoType field, by checking that it implements the GQL marshaller interfaces.

giautm: Are all types implements the GQL marshaller interfaces is a scalar? How we can check for UUID case? It's not implement GQL marshaller interfaces, but using custom marshal functions
```
  UUID:
    model:
      - entgo.io/contrib/entgql/internal/todouuid/ent/schema/uuidgql.UUID
```
 This allow user defined their custom scalar type with more information

entgql/extension.go
a8m: Got it, agree.

entgql/extension.go
a8m: Got it, agree.

giautm:
Let this part be handled later

PR #256, opened 2022-03-08
entgql: generate edge fields

giautm:
Related to this comment: https://github.com/ent/ent/issues/2351#issuecomment-1057135721

### DIFF
--- a/entgql/plugin.go
+++ b/entgql/plugin.go
@@ -64,7 +64,7 @@ func (e *gqlgenPlugin) InjectSourceEarly() *ast.Source {
 func (e *gqlgenPlugin) MutateConfig(cfg *config.Config) error {
 	if e.models != nil {
 		for name, goType := range e.models {
-			if len(cfg.Models[name].Model) == 0 {
+			if !cfg.Models.UserDefined(name) {
 				cfg.Models.Add(name, goType)
 			}
 		}
--- a/entgql/plugin.go
+++ b/entgql/plugin.go
@@ -64,7 +64,7 @@ func (e *gqlgenPlugin) InjectSourceEarly() *ast.Source {
 func (e *gqlgenPlugin) MutateConfig(cfg *config.Config) error {
 	if e.models != nil {
 		for name, goType := range e.models {
-			if !cfg.Models.Exists(name) {
+			if len(cfg.Models[name].Model) == 0 {
 				cfg.Models.Add(name, goType)
 			}
 		}
--- a/entgql/schema.go
+++ b/entgql/schema.go
@@ -197,12 +197,12 @@ func (e *schemaGenerator) buildTypes(types map[string]*ast.Definition) error {
 			if ant.RelayConnection && edge.Unique {
 				return fmt.Errorf("RelayConnection cannot be defined on Unique edge: %s.%s", node.Name, edge.Name)
 			}
-			field, err := e.buildEdge(edge, ant)
+			fields, err := e.buildEdge(edge, ant)
 			if err != nil {
 				return err
 			}
-			if field != nil {
-				typ.Fields = append(typ.Fields, field)
+			if len(fields) > 0 {
+				typ.Fields = append(typ.Fields, fields...)
 			}
 		}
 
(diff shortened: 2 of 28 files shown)


entgql/plugin.go
```diff
@@ -64,7 +64,7 @@ func (e *gqlgenPlugin) InjectSourceEarly() *ast.Source {
 func (e *gqlgenPlugin) MutateConfig(cfg *config.Config) error {
 	if e.models != nil {
 		for name, goType := range e.models {
-			if !cfg.Models.Exists(name) {
```
giautm, line 67: This fixed issue when we use the `@goField` directive on the field.
[It creates a mapping without Models](https://github.com/99designs/gqlgen/blob/f0fdb116f45350aabf698c20bf6410283f96bb11/codegen/config/config.go#L285-L290). Then, we don't have the correct map so GQLGen will generate models for it.

It's maybe related to this issue: https://github.com/99designs/gqlgen/discussions/1993

https://github.com/ent/contrib/pull/78/files#diff-14325d4666b9e54a422296f16c74f5d5f155958fecba87f9144b511f92b23b4dR47-R49

cc @cliedeman

entgql/plugin.go
```diff
@@ -64,7 +64,7 @@ func (e *gqlgenPlugin) InjectSourceEarly() *ast.Source {
...
+			if len(cfg.Models[name].Model) == 0 {
```
giautm, line 67: This fixed issue when we use the `@goField` directive on the field.
[It creates a mapping without Models](https://github.com/99designs/gqlgen/blob/f0fdb116f45350aabf698c20bf6410283f96bb11/codegen/config/config.go#L285-L290). Then, we don't have the correct map so GQLGen will generate models for it.

It's maybe related to this issue: https://github.com/99designs/gqlgen/discussions/1993

https://github.com/cliedeman/entcontrib/blob/60b59d6c5a2656e90f6b78f9ba88645f17b60b35/entgql/plugin/config.go#L47-L49

cc @cliedeman

entgql/schema.go
```diff
@@ -319,6 +328,59 @@ func (e *schemaGenerator) buildTypeFields(t *gen.Type) (ast.FieldList, error) {
...
+		return nil, err
+	}
+
+	edgeField := camel(edge.Name)
+	var mappings []string
+	if len(edgeAnt.Mapping) > 0 {
+		mappings = edgeAnt.Mapping
+	} else {
+		mappings = []string{edgeField}
+	}
```
a8m, line 343:
```suggestion
	var (
		edgeField = camel(edge.Name)
		mappings = []string{edgeField}
	)
	if len(edgeAnt.Mapping) > 0 {
		mappings = edgeAnt.Mapping
	}
```

entgql/schema.go
```diff
@@ -319,6 +328,59 @@ func (e *schemaGenerator) buildTypeFields(t *gen.Type) (ast.FieldList, error) {
...
+		switch {
+		case edge.Unique:
+			fieldDef.Type = namedType(gqlType, edge.Optional)
+		case edgeAnt.RelayConnection:
+			if !e.relaySpec {
+				return nil, ErrRelaySpecDisabled
+			}
+			if !ant.RelayConnection {
+				return nil, fmt.Errorf("entgql: must enable Relay Connection via the entgql.RelayConnection annotation on the %s entity", edge.Type.Name)
+			}
```
a8m, line 363: qq about this - in case a Node is marked as `RelayConnection` in its schema, do we add to the Query object?

entgql/schema_test.go
```diff
@@ -80,6 +81,9 @@ type Todo {
   status: Status!
   priority: Int!
   text: String!
+  parent: Todo
+  childrenConnection: [Todo!] @goField(name: "children", forceResolver: false)
```
a8m, line 85: Is it a wrong naming or a bug?

entgql/template/pagination.tmpl
```diff
@@ -280,8 +280,10 @@ func (c *{{ $conn }}) build(nodes []*{{ $name }}, pager *{{ $pager }}, first, la
 			Cursor: pager.toCursor(node),
 		}
 	}
-	c.PageInfo.StartCursor = &c.Edges[0].Cursor
-	c.PageInfo.EndCursor = &c.Edges[len(c.Edges)-1].Cursor
+	if l := len(c.Edges); l > 0 {
+		c.PageInfo.StartCursor = &c.Edges[0].Cursor
+		c.PageInfo.EndCursor = &c.Edges[l-1].Cursor
+	}
```
a8m, line 286: 💯

a8m (approved):
Looks good. A few minor comments

entgql/schema_test.go
giautm: Its just wrong naming in the schema then it effect the test, isnot a bug

PR #257, opened 2022-03-10
feat(entgql/pagination): support custom GoTypes for string ID fields

tarrencev:
Implements support for polymorphic id go types for string ids. When using GoTypes, `entgql` will attempt to use the GQLMarshaler interface to normalize them to the graphql schema type.

Currently, we try to infer the id type from nodes (im not sure if there is a reason for this?) this PR changes it to use the explicitly default type defined by `gen.Config`.


entgql/template/node.tmpl
```diff
@@ -15,13 +15,13 @@ in the LICENSE file in the root directory of this source tree.
 {{ end }}
 
 {{ $gqlNodes := filterNodes $.Nodes }}
-{{ $idType := findIDType $gqlNodes $.IDType }}
+{{ $v := assertIDTypes $gqlNodes $.IDType }}
```
tarrencev, line 18: is there a better way to handle this? `assertIDTypes` could just return an error but then it outputs a `<nil>` into the generate code

giautm:
Hello, please split your commit to smaller chunks so we can review carefully.

I dont think you should change behavior of findIDType to assertType. Because all most user currently not setting it. On the other hand, some user just want to expose some parts of the schema to GQL. And the ID cannot be same the global config.

tarrencev:
> Hello, please split your commit to smaller chunks so we can review carefully.
> 
> I dont think you should change behavior of findIDType to assertType. Because all most user currently not setting it. On the other hand, some user just want to expose some parts of the schema to GQL. And the ID cannot be same the global config.

@giautum i can rewrite the findIDType to infer it again. im not sure it is possible to make this much smaller though. the actual change is like 100 lines. everything else is generated

giautm:
Just split to at least two commit: one for generated, other for your work.

giautm:
> Implements support for polymorphic id go types for string ids

Why do you need this feature? What's issue are you trying to resolve?

tarrencev:
The todogotype provides an example. With ent it is possible to have custom gotypes for ids, as long as they implement driver.Value. Graphql requires homogenous id types. Before we were naively assuming all id types were the same. This normalizes id types when the ent schema has multiple id gotypes

giautm:
Hey, thanks for sharing the intent.

Let me think about my own implementation for this issue in a few days. Then, I will be back to review your changes again to make sure we choose the right approach for it.

tarrencev:
ok thanks. looking forward to your feedback

tarrencev:
hey @giautm just wanted to ping this to see if you have any thoughts?

giautm:
Hey @tarrencev, sorry for the delay. I'm working on it right now.

The only thing I consider is the term `polymorphic`, It's not a good term for GQL/Ent.

I think we can officially support for GQLGen Marshal/Unmarshal as you did. But some changes are needed. I will update you quickly. :D 

PS: If you are blocked by this, just use Go modules replace to point EntGQL to your fork.

tarrencev:
Hi @giautm, thanks for getting back to me. I agree `polymorphic` is a bad word, perhaps we can call it "id normalization" or something? Glad you think we can officially support it and excited to see your solution. Feel free to take over the PR if you would like!

giautm:
I really appreciate your contribution, 🙇 

If this PR is released, we will be able to update the examples for UUID or ULID in a simpler way.

giautm:
Hey, I opened PR #262 with my updates.

The most [important change](https://github.com/ent/contrib/pull/262/commits/1d4dca43bea720ff6fbfd3294584a8c047643ab7) is convert `polymorphic` guess to `WithStringID()` option. I think this could be the correct name for this feature. Because It's always required every Entity has an ID field with the String type. Later, the template uses Marshal/Unmarshal functions to convert the ID from/to string type.

Followup question: Is must be to use MarsharlGQL/UnmarsharlGQL interfaces? Or MarshalText/UnmarshalText interface?

giautm:
> One challenge i had was how to use the gql marshaller for a custom third part type

We intend to resolve this problem when we implement the EntGQL plugin. The plugin may provide an option as GQLGen did to map marshal/unmarshal functions.

tarrencev:
> Followup question: Is must be to use MarsharlGQL/UnmarsharlGQL interfaces? Or MarshalText/UnmarshalText interface?

I was thinking `MarshalGQL` made most sense since it needs to be implemented anyways to support the id types, but would be happy either way. I was actually thinking ideally we could use `driver.Valuer`, but couldn't figure out a good way to do support that

giautm:
I have updated the master branch to use the latest version of ent. So, this PR known only have the changes from our work.
- I will continue to test it with the WhereInput feature to make sure it doesn't break anything.
- I will be trying to upgrade the PULID to use this feature because it looks promising with these changes.

tarrencev:
hey @giautm just wanted to check in on this :)

giautm:
Hello @tarrencev, I'm still following this. But need to finish the current work first. There are conflicts that will arise if this is done.

tarrencev:
> Hello @tarrencev, I'm still following this. But need to finish the current work first. There are conflicts that will arise if this is done.

Ok sounds good. Do you think the api will look similar/is safe to start using? Just want to be careful not to back ourselves into a corner

giautm (approved):
LGTM, because I touched it. 🙃

a8m:
Closed with https://github.com/ent/contrib/pull/302. 

Thanks @tarrencev and @giautm

PR #260, opened 2022-03-17
added field annotation for read only flag

rolandknight:
Added FieldConfig struct to Annotation to support read only fields. Corresponding change coming to ogent that uses this flag to suppress parameter generation for read only fields (usefult for createdBy and createdAt fields).

### DIFF
--- a/entoas/annotation.go
+++ b/entoas/annotation.go
@@ -32,8 +32,6 @@ type (
 		Example interface{}
 		// OpenAPI Specification schema to use for a schema field.
 		Schema *ogen.Schema
-		// FieldConfig has meta information about a schema field.
-		Field FieldConfig
 		// Create has meta information about a creation operation.
 		Create OperationConfig
 		// Read has meta information about a read operation.
@@ -44,6 +42,8 @@ type (
 		Delete OperationConfig
 		// List has meta information about a list operation.
 		List OperationConfig
+		// ReadOnly specifies that the field/edge is read only (no create/update parameter)
+		ReadOnly bool
 	}
 	// OperationConfig holds meta information about a REST operation.
 	OperationConfig struct {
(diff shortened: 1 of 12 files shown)


entoas/annotation.go
```diff
@@ -50,6 +52,10 @@ type (
 	}
 	// OperationConfigOption allows managing OperationConfig using functional arguments.
 	OperationConfigOption func(*OperationConfig)
+	// FieldConfig hold meta information aboat a REST field (parameter)
+	FieldConfig struct {
+		ReadOnly bool
+	}
```
masseelch, line 58: I think we can drop this and add the `ReadOnly` field directly on the `Annotation`.

masseelch (changes requested):
Hey @rolandknight, thank you for this!

Can you please address my comment and add a test for the use case?

entoas/annotation.go
rolandknight: Done. 1. ReadOnly directly in Annotation, 2. Builder added, 3. Unit tests added

Once merged, I'll update the corresponding ogent PR

entoas/annotation.go
```diff
@@ -44,6 +42,8 @@ type (
 		Delete OperationConfig
 		// List has meta information about a list operation.
 		List OperationConfig
+		// ReadOnly specifies that the field/edge is read only (no create/update parameter)
```
masseelch, line 45:
```suggestion
		// ReadOnly specifies that the field/edge is read only (no create/update parameter).
```

entoas/annotation.go
```diff
@@ -104,6 +100,11 @@ func ListOperation(opts ...OperationConfigOption) Annotation {
 	return Annotation{List: operationsConfig(opts)}
 }
 
+// ReadOnly returns a read only field/edge annotation
```
masseelch, line 103:
```suggestion
// ReadOnly returns a read only annotation.
```

masseelch (changes requested):
Thanks! One thing left to do is to add the `ReadOnly` property to the `Merge()` method.

rotemtam:
Can't we use .Immutable() here ?

masseelch:
> Can't we use .Immutable() here ?

No. Because fields like `created_at` are not immutable. But they should not be part of the generated POST body when attempting to create a new schema object.

masseelch (approved):
:100: Very well done and thank you very much :blush:

PR #267, opened 2022-03-23
entgql(WhereInput): remove graphql-go/graphql

giautm:
We replace graphql-go/graphql with vektah/gqlparser

~This commit includes small break changes, that is will not
update the existed input in the schema.
You HAVE TO remove the duplicated input from your schema.~

Refactor WhereInput feature for https://github.com/ent/ent/issues/2351


giautm:
BC: always overwritten `ent.gql`

In the previous version, the EntGQL extension will check if the input data type exists in the `ent.gql` file, it will be left as is.
With this version, all operations are always-overridden. Therefore, if you declare the input data type in the `ent.gql` file, it will be deleted and replaced with the version of EntGQL.

entgql/extension.go
```diff
@@ -378,72 +356,39 @@ func (e *Extension) whereExists() (int, bool) {
...
-				// If the input object was found in the schema, we update its definition.
-				if node, ok := p.Node.(*ast.InputObjectDefinition); ok && inputs[node.Name.Value] != nil {
-					input := inputs[node.Name.Value]
-					delete(inputs, node.Name.Value)
-					return visitor.ActionUpdate, input
-				}
-				return visitor.ActionNoChange, nil
-			},
-		},
-	}, nil)
```
giautm, line 394: This's causes breaking changes. We can restore the old behavior but it required to parse the `ent.gql` file with `vektah/gqlparser`

crossworth:
@giautm do you think it would be possible to add support for some kind of "hook" on the whereFilters part?

Maybe allowing us (users) to provide a custom `genWhereInputs` and export the default one (and use it when not provided), maybe export the others methods related to it as well (so we can use they on our genWhereInputs user-side implementation).

Right now, even if we change the templates for the where filters, there is no way to change the `ent.graphql` file, I need to add custom filters/directives on the generated file.

giautm:
Hey @crossworth, thank you for your kind words.

I plan to add some hooks for modifying the schema after we generate it from the graph. But, currently, the WhereInput and GQL Schema aren't connected yet to become a single schema. I will continue on it after this PR get merged.

entgql/extension.go
```diff
@@ -30,29 +30,20 @@ import (
...
-	"github.com/graphql-go/graphql"
-	"github.com/graphql-go/graphql/language/ast"
-	"github.com/graphql-go/graphql/language/kinds"
-	"github.com/graphql-go/graphql/language/parser"
-	"github.com/graphql-go/graphql/language/printer"
-	"github.com/graphql-go/graphql/language/source"
-	"github.com/graphql-go/graphql/language/visitor"
-
-	ast2 "github.com/vektah/gqlparser/v2/ast"
+	"github.com/vektah/gqlparser/v2/ast"
```
a8m, line 33: ❤️

entgql/extension.go
a8m: This logic was used to allow extending an existing GraphQL schema with the generated `WhereInput`s. Do we want to drop this support? We don't use the "upsert" option. 

cc @crossworth

entgql/internal/todo/ent.graphql
```diff
@@ -3,151 +3,165 @@ CategoryWhereInput is used for filtering Category objects.
...
+	idGT: ID
+	idGTE: ID
+	idLT: ID
+	idLTE: ID
+	"""
+	todos edge predicates
+	"""
+	hasTodos: Boolean
+	hasTodosWith: [TodoWhereInput!]
 }
```
a8m, line 87: Can we control the printer to use 2 spaces instead of a tab? I think it's more common in GraphQL (and JS).

a8m (commented):
I love it, just a few comments before accepting. 

Great work @giautm.

entgql/internal/todo/ent.graphql
giautm: Nope, there's no such option in the gqlparser's formatter.

entgql/extension.go
giautm: As I mentioned in [this comment](https://github.com/ent/contrib/pull/267#issuecomment-1077936733), I will support the hook mechanism to update the GQL schema.

entgql/internal/todo/ent.graphql
giautm: Sent PR: https://github.com/vektah/gqlparser/pull/208

a8m: 💯

a8m (approved):
Great work 💯

a8m:
Thanks for landing this @giautm. For next time, please use `<package>: <message>` as a commit format.

Also, your PRs were merged to `gqlparser`, so you're welcome to add the printer change here 😃

PR #268, opened 2022-03-24
entgql: refactor Skip annotation

giautm:
Related to https://github.com/ent/ent/issues/2351#issuecomment-1057315202

### DIFF
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -16,7 +16,6 @@ package entgql
 
 import (
 	"encoding/json"
-	"fmt"
 
 	"entgo.io/ent/entc/gen"
 	"entgo.io/ent/schema"
@@ -70,8 +69,6 @@ const (
 	SkipType SkipMode = 1 << iota
 	// SkipEnumField will skip generating the enum type from the enum field
 	SkipEnumField
-	// SkipMutationInput will skip generating the entity or the field in the MutationInput
-	SkipMutationInput
 	// SkipOrderField will skip generating the entity or the field for the enum order
 	SkipOrderField
 	// SkipWhereInput will skip generating the entity or the field in the WhereInput
@@ -80,7 +77,6 @@ const (
 	// SkipAll is default mode to skip all
 	SkipAll = SkipType |
 		SkipEnumField |
-		SkipMutationInput |
 		SkipOrderField |
 		SkipWhereInput
 )
@@ -213,23 +209,6 @@ func (f SkipMode) Any() bool {
 // Has check if the skip annotation has a specific flag
 func (f SkipMode) Has(check SkipMode) bool {
 	return f&check == check
-}
-
-// skipModeFromString returns SkipFlag from a string
-func skipModeFromString(s string) (SkipMode, error) {
-	switch s {
-	case "type":
-		return SkipType, nil
-	case "enum_field":
-		return SkipEnumField, nil
-	case "order_field":
-		return SkipOrderField, nil
-	case "mutation_input":
-		return SkipMutationInput, nil
-	case "where_input":
-		return SkipWhereInput, nil
-	}
-	return 0, fmt.Errorf("invalid skip mode: %s", s)
 }
 
 // annotation extracts the entgql.Annotation or returns its empty value.
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -65,16 +65,17 @@ type (
 )
 
 const (
-	// SkipType will skip generating the entity or the field in the schema
+	// SkipType skips generating GraphQL types or fields in the schema.
 	SkipType SkipMode = 1 << iota
-	// SkipEnumField will skip generating the enum type from the enum field
+	// SkipEnumField skips generating GraphQL enums for enum fields in the schema.
 	SkipEnumField
-	// SkipOrderField will skip generating the entity or the field for the enum order
+	// SkipOrderField skips generating GraphQL order inputs and enums for ordered-fields in the schema.
 	SkipOrderField
-	// SkipWhereInput will skip generating the entity or the field in the WhereInput
+	// SkipWhereInput skips generating GraphQL WhereInput types.
+	// If defined on a field, the type will be generated without the field.
 	SkipWhereInput
 
-	// SkipAll is default mode to skip all
+	// SkipAll is default mode to skip all.
 	SkipAll = SkipType |
 		SkipEnumField |
 		SkipOrderField |
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -209,7 +209,7 @@ func (f SkipMode) Any() bool {
 
 // Is checks if the skip annotation has a specific flag.
 func (f SkipMode) Is(mode SkipMode) bool {
-	return f&mode != 0 
+	return f&mode != 0
 }
 
 // annotation extracts the entgql.Annotation or returns its empty value.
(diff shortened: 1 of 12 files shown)


entgql/annotation.go
```diff
@@ -59,6 +60,29 @@ type (
 		Value string        `json:"value,omitempty"`
 		Kind  ast.ValueKind `json:"kind,omitempty"`
 	}
+
+	// SkipFlag is a bit flag for the skip annotation .
+	SkipFlag int
```
a8m, line 65: SkipMode is a better name imo.

entgql/annotation.go
```diff
@@ -59,6 +60,29 @@ type (
...
+	// SkipFlagType will skip generate the entity/field in the query
+	SkipFlagType SkipFlag = 1 << iota
+	// SkipFlagFieldEnum will skip generate Enum from the enum field
+	SkipFlagFieldEnum
+	// SkipFlagInput will skip mutation input for the entity/field
+	SkipFlagInput
+	// SkipFlagWhere will skip input filter for the entity/field
+	SkipFlagWhere
+	// SkipFlagOrder will skip generate sort order for the entity
+	SkipFlagOrder
```
a8m, line 78:
```suggestion
	SkipNode SkipFlag = 1 << iota
	// SkipEnum will skip generate Enum from the enum field
	SkipEnumField
	// SkipMutationInput will skip mutation input for the entity/field
	SkipMutationInput
	// SkipWhereInput will skip input filter for the entity/field
	SkipWhereInput
	// SkipOrderField will skip generate sort order for the entity
	SkipOrderField
```

I'm thinking about how the API will look and I shorter feels cleaner to me. Another point, what does it mean to `SkipOrderField` or `SkipEnumField`? Skip generating only GraphQL or also in Ent? Do we need a flag for both options?

giautm: `SkipNode`: I thought about this name before, because we need to use it for all fields. So `SkipNode` does not seem appropriate. Do we need to add `SkipField` as well? `SkipType` seems more suitable to me, as it comes with `SkipInput`


> Another point, what does it mean to SkipOrderField or SkipEnumField? Skip generating only GraphQL or also in Ent

It should skip generating both Go code and GQL schema.

a8m: > `SkipType` seems more suitable to me, as it comes with `SkipInput`

Right, let's use `SkipType`. I wonder how's the usage will look like:

```go
func (T) Annotations() []schema.Annotation {
    return []schema.Annotation{
        // Default to skip All.
        entgql.Skip(),

        // Or, skip only type and input generation.
        entgql.Skip(entgql.SkipType|entgql.SkipInput),
    }
}

func (T) Fields() []ent.Field {
    return []ent.Field{
        field.Int("age").
            Annotations(
                // Skip all.
                entgql.Skip(),

                // Or, skip only from input?
                entgql.Skip(entgql.SkipInput),
            ),
    }
}
```

This is the API you thought about or you have something else in mind?

giautm: Yes, it's almost the same.

```go
func (T) Annotations() []schema.Annotation {
    return []schema.Annotation{
        // Default to skip All.
        entgql.Skip(),

        // Or, skip only type and input generation.
        entgql.Skip(entgql.SkipType, entgql.SkipMutationInput),
    }
}

func (T) Fields() []ent.Field {
    return []ent.Field{
        field.Int("age").
            Annotations(
                // Skip all.
                entgql.Skip(),

                // Or, skip only from input?
                entgql.Skip(entgql.SkipMutationInput),
            ),
    }
}
```

entgql/annotation.go
```diff
@@ -59,6 +59,26 @@ type (
...
+	// SkipOrderField will skip generating the entity or the field for the enum order
+	SkipOrderField
+	// SkipWhereInput will skip generating the entity or the field in the WhereInput
+	SkipWhereInput
+
+	// SkipAll is default mode to skip all
+	SkipAll = SkipType |
+		SkipEnumField |
+		SkipOrderField |
+		SkipWhereInput
```
giautm, line 81: Hey @a8m, let's review the naming and the description again.

entgql/annotation.go
```diff
@@ -59,6 +59,26 @@ type (
...
 )
```
a8m, line 82:
```suggestion
const (
	// SkipType skips generating GraphQL types or fields in the schema.
	SkipType SkipMode = 1 << iota
	// SkipEnumField skips generating GraphQL enums for enum fields in the schema.
	SkipEnumField
	// SkipOrderField skips generating GraphQL order inputs and enums for ordered-fields in the schema.
	SkipOrderField
	// SkipWhereInput skips generating GraphQL WhereInput types.
	// If defined on a field, the type will be generated without the field.
	SkipWhereInput

	// SkipAll is default mode to skip all.
	SkipAll = SkipType |
		SkipEnumField |
		SkipOrderField |
		SkipWhereInput
)
```

entgql/annotation.go
```diff
@@ -173,6 +201,16 @@ func (a *Annotation) Decode(annotation interface{}) error {
 	return json.Unmarshal(buf, a)
 }
 
+// Any returns true if the skip annotation is setted
+func (f SkipMode) Any() bool {
+	return f != 0
+}
```
a8m, line 207: Why it's preferred over `!= 0`?

entgql/annotation.go
```diff
@@ -173,6 +201,16 @@ func (a *Annotation) Decode(annotation interface{}) error {
 	return json.Unmarshal(buf, a)
 }
 
+// Any returns true if the skip annotation is setted
```
a8m, line 204:
```suggestion
// Any returns true if the skip annotation was set.
```

entgql/annotation.go
```diff
@@ -173,6 +201,16 @@ func (a *Annotation) Decode(annotation interface{}) error {
...
+func (f SkipMode) Any() bool {
+	return f != 0
+}
+
+// Has check if the skip annotation has a specific flag
+func (f SkipMode) Has(check SkipMode) bool {
+	return f&check == check
+}
```
a8m, line 212:
```suggestion
// Is checks if the skip annotation has a specific flag.
func (f SkipMode) Is(mode SkipMode) bool {
	return f&mode != 0 
}
```

ent uses `Is` in multiple places for big flags.

entgql/template.go
```diff
@@ -71,6 +71,8 @@ var (
 		"filterNodes":         filterNodes,
 		"findIDType":          findIDType,
 		"nodePaginationNames": nodePaginationNames,
+		"skipMode":            skipModeFromString,
+		"hasSkipMode":         hasSkipMode,
```
a8m, line 75: This function is not needed IMO, you can use the `skipMode` function in the template (0 is falsy).

entgql/template.go
```diff
@@ -152,15 +154,15 @@ func filterNodes(nodes []*gen.Type) ([]*gen.Type, error) {
 }
 
 // filterEdges filters out edges that should not be included in the GraphQL schema.
-func filterEdges(edges []*gen.Edge) ([]*gen.Edge, error) {
+func filterEdges(edges []*gen.Edge, skip SkipMode) ([]*gen.Edge, error) {
```
a8m, line 157: Make it an optional param to not break existing templates used by users.

a8m (approved):

PR #269, opened 2022-03-24
entgql: initial work for adding nested pagination

a8m:
This is an initial work that generates a paginate function for each edge `E` with the `RelayConnection` annotation.

This is not useful by itself, because there are 2 additional PRs that need to come after it for loading the edges using eager-loading and extracting the cached result on this function. However, some refactoring complexity was added to this PR to make the review of future PRs easier.


giautm (approved):
LGTM, let me continue working on https://github.com/ent/contrib/pull/256 after this get merge

a8m:
> LGTM, let me continue working on #256 after this get merge

Just note, I have additional PRs that are rebased on this. I'll you know when I'm done.

PR #271, opened 2022-03-25
entoas: added two new parameter to filter and sort a list of models.

mauriciobento:
Added two new parameters for API. OrderBy and filter.
The value of the **orderBy** parameter contains a comma-separated list of expressions used to sort the items. A special case of such an expression is a property path terminating on a primitive property.
The **filter** querystring parameter allows clients to filter a collection of resources that are addressed by a request URL


entoas/generator.go
```diff
@@ -432,6 +432,16 @@ func listOp(spec *ogen.Spec, n *gen.Type) (*ogen.Operation, error) {
 				SetName("itemsPerPage").
 				SetDescription("item count to render per page").
 				SetSchema(ogen.Int()),
+			ogen.NewParameter().
+				InQuery().
+				SetName("orderBy").
+				SetDescription("clause is used to sort the records in the result set for a SELECT statement.").
```
masseelch, line 438:
```suggestion
				SetDescription("clause is used to sort the records in the result set for a SELECT statement").
```

entoas/generator.go
```diff
@@ -478,6 +488,16 @@ func listEdgeOp(spec *ogen.Spec, n *gen.Type, e *gen.Edge) (*ogen.Operation, err
...
```
masseelch, line 494:
```suggestion
				SetDescription("clause is used to sort the records in the result set for a SELECT statement").
```

masseelch (changes requested):
Hey @mauriciobento! Thanks for this awesome contribution. I will now have a look at it.

masseelch:
Closing, since we are not merging this feature as discussed in https://github.com/ariga/ogent/pull/74

PR #273, opened 2022-03-26
entgql: move the order-fields logic to template function

a8m:
If the target `Edge.Type` does not have order fields, they won't be defined in the GraphQL schema, and having them in the method signature will break the gqlgen binding.

### DIFF
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -164,7 +164,7 @@ func (a Annotation) Merge(other schema.Annotation) schema.Annotation {
 	return a
 }
 
-// Decode unmarshal annotation
+// Decode unmarshalls the annotation.
 func (a *Annotation) Decode(annotation interface{}) error {
 	buf, err := json.Marshal(annotation)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


entgql/template.go
```diff
@@ -261,3 +264,14 @@ func removeOldTemplate(g *gen.Graph, name string) error {
 	}
 	return nil
 }
+
+// annotation extracts the entgql.Annotation or returns its empty value.
+func annotation(ants gen.Annotations) (*Annotation, error) {
```
giautm, line 269: please use `decodeAnnotation` function,

a8m: decode feels to me like an internal detail. From the caller's perspective, I'm interested to get the `entgql.Annotation` and don't care how it's done.

giautm: That function is existes in the codebase, so re-use it or rename it, please.

a8m: oh, got it. I didn't see the `decodeAnnotation` function. Thanks

entgql/annotation.go
```diff
@@ -173,15 +173,13 @@ func (a *Annotation) Decode(annotation interface{}) error {
...
+// annotation extracts the entgql.Annotation or returns its empty value.
+func annotation(ants gen.Annotations) (*Annotation, error) {
 	ant := &Annotation{}
-	if annotations == nil || annotations[ant.Name()] == nil {
-		return ant, nil
-	}
-
-	if err := ant.Decode(annotations[ant.Name()]); err != nil {
-		return nil, err
+	if ants != nil || ants[ant.Name()] != nil {
```
giautm, line 179:
```suggestion
	if ants != nil && ants[ant.Name()] != nil {
```

giautm (changes requested):
I think there is a mistake

entgql/annotation.go
a8m: Oops. Thanks @giautm

giautm (approved):
LGTM, <3

PR #277, opened 2022-03-31
entgql: initial work for nested pagination in templates

a8m:
PR is ready for review. Implementation examples exist in entgql/internal/todo/todo_test.go#926.

Additional parts that should be added after this PR:
- It's better IMO, to not rely on gqlgen to unmarshal and load the arguments and move the (un)marshling logic of `<T>WhereInput` types to ent. 
- Add additional tests/examples and publish a tutorial or a blog post about it?
- The code for loading the `totalCount` can be improved by removing some of the duplications there.


go.mod
```diff
@@ -2,6 +2,10 @@ module entgo.io/contrib
 
 go 1.18
 
+replace github.com/99designs/gqlgen => ../gqlgen
+
+replace entgo.io/ent => ../ent
+
```
masseelch, line 8: remove

a8m: removed

entgql/template.go
```diff
@@ -102,35 +103,28 @@ func findIDType(nodes []*gen.Type, defaultType *field.TypeInfo) (*field.TypeInfo
...
 		if err != nil {
 			return nil, err
 		}
-		if ant.Unbind {
-			delete(result, e.Name)
-		}
-		if len(ant.Mapping) > 0 {
-			if _, bind := result[e.Name]; bind {
+		switch {
+		case len(ant.Mapping) > 0:
```
giautm, line 118: 💯

giautm (approved):
LGTM, but we should upgrade ent/gqlgen in the separated commit to avoid changes in other packages.

PR #281, opened 2022-04-10
entgql: refactor schema

giautm:


entgql/internal/todoplugin/gqlgen.yml
```diff
@@ -21,7 +21,11 @@ models:
   Duration:
     model:
       - entgo.io/contrib/entgql/internal/todo/ent/schema/durationgql.Duration
-  # Custom mapping from GraphQL `CategoryConfigInput` to `schematype.CategoryConfig`.
+  # Custom mapping from GraphQL `CategoryConfig` and
+  # `CategoryConfigInput` to `schematype.CategoryConfig`.
+  CategoryConfig:
+    model:
+      - entgo.io/contrib/entgql/internal/todo/ent/schema/schematype.CategoryConfig
```
giautm, line 28: I revert the support for custom Go types because the GQL type was not defined by Ent. The user should manual mapping for it the gqlgen's config or use `@goModel` to GQL type.

entgql/schema.go
```diff
@@ -527,15 +528,15 @@ func (e *schemaGenerator) entGoType(name string) string {
...
+	if f.IsEnum() {
+		if f.HasGoType() {
+			return fmt.Sprintf("%s.%s", f.Type.RType.PkgPath, f.Type.RType.Name), true
+		}
 		return fmt.Sprintf("%s/%s", e.graph.Package, f.Type.Ident), true
-	default:
-		return "", false
 	}
+
+	return "", false
```
giautm, line 539: Only mapping for the enum types

entgql/internal/todoplugin/gqlgen.yml
a8m: Can’t we auto inject the goModel directive?

entgql/schema.go
```diff
@@ -486,7 +486,8 @@ func (e *schemaGenerator) genModels() (map[string]string, error) {
 				hasOrderBy = true
 			}
 
-			goType, ok := e.fieldGoType(field)
+			// We only map the Go types generated by Ent, for example: enum
```
a8m, line 489:
```suggestion
			// We only map the Go types generated by Ent, for example: enum.
```

a8m (approved):
💯

entgql/internal/todoplugin/gqlgen.yml
giautm: It can be done by using the plugin only. But when we write the output to the file, we should not modify user-defined files

PR #283, opened 2022-04-11
feat(entoas): Support RequestGroups to determine what fields to serialize for the requestBody

omarhachach:
If a better way to do it is available, feel free to let me know 😄


omarhachach:
Do note: using RequestGroups on edges breaks ogent: https://github.com/ariga/ogent/blob/main/template/update.tmpl#L17
This is due to ogent assuming that all edges are present always, which they wont be anymore.

entoas/annotation.go
```diff
@@ -64,6 +66,11 @@ func OperationGroups(gs ...string) OperationConfigOption {
 	return func(c *OperationConfig) { c.Groups = gs }
 }
 
+// RequestGroups returns a OperationConfigOption that adds the given request serialization groups to a OperationConfig.
+func RequestGroups(gs ...string) OperationConfigOption {
```
masseelch, line 70:
```suggestion
func OperationRequestGroups(gs ...string) OperationConfigOption {
```

entoas/annotation_test.go
```diff
@@ -17,10 +17,11 @@ package entoas
 import (
 	"testing"
 
-	"entgo.io/contrib/entoas/serialization"
 	"entgo.io/ent/entc/gen"
 	"github.com/ogen-go/ogen"
 	"github.com/stretchr/testify/require"
+
+	"entgo.io/contrib/entoas/serialization"
```
masseelch, line 24: dito

masseelch (commented):
Some minor comments :-)

entoas/annotation.go
omarhachach: Indeed better 👍🏽

masseelch (approved):
:100:

entoas/view.go
```diff
@@ -170,10 +171,10 @@ func serializeField(f *gen.Field, g serialization.Groups) (bool, error) {
 }
 
 // serializeEdge checks if an edge is to be serialized according to its annotations and the requested groups.
-func serializeEdge(e *gen.Edge, g serialization.Groups) (bool, error) {
+func serializeEdge(e *gen.Edge, g serialization.Groups, includeNoGroup bool) (bool, error) {
 	// If no groups are requested or the edge has no groups defined do not render the edge.
 	if e.Annotations == nil || len(g) == 0 {
-		return false, nil
+		return includeNoGroup, nil
```
omarhachach, line 177: This is done so the behaviour matches the previous default behaviour

PR #285, opened 2022-04-18
entproto: support field.Strings

rotemtam:
Closes https://github.com/ent/ent/issues/2474


entproto/adapter.go
```diff
@@ -510,9 +515,20 @@ func extractProtoTypeDetails(f *gen.Field) (fieldType, error) {
 	}, nil
 }
 
+func extractJSONDetails(f *gen.Field) (fieldType, error) {
+	if f.Type.Ident == "[]string" {
```
giautm, line 519: We also have []float64 and []int

rotemtam: Yep, this PR adds support for field.Strings, the rest will come after

entproto/internal/entprototest/adapter_test.go
```diff
@@ -94,6 +94,14 @@ func (suite *AdapterTestSuite) TestImplicitSkippedMessage() {
 	suite.EqualError(err, entproto.ErrSchemaSkipped.Error())
 }
 
+func (suite *AdapterTestSuite) TestMessageWithStrings() {
```
a8m, line 97: Do you mind changing the receiver to be a single character? In this PR or a different one. Thanks 🙏

a8m (approved):

entproto/internal/entprototest/adapter_test.go
rotemtam: Sure. I want to deprecate the usage of suite anyways

PR #289, opened 2022-04-21
entgql: expose the Entity connection on the Query type

giautm:
Related issue: https://github.com/ent/ent/issues/2351

### DIFF
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -46,7 +46,7 @@ type (
 		// Directives to add on the field/type.
 		Directives []Directive `json:"Directives,omitempty"`
 		// QueryField exposes the generated type with the given string under the Query object.
-		QueryField *QueryFieldConfig `json:"QueryField,omitempty"`
+		QueryField *FieldConfig `json:"QueryField,omitempty"`
 	}
 
 	// Directive to apply on the field/type
@@ -65,7 +65,7 @@ type (
 	// SkipMode is a bit flag for the Skip annotation.
 	SkipMode int
 
-	QueryFieldConfig struct {
+	FieldConfig struct {
 		// Name is the name of the field in the Query object.
 		Name string `json:"Name,omitempty"`
 		// Directives to add on the field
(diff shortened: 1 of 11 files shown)


go.sum
```diff
@@ -147,7 +147,6 @@ github.com/mattn/go-isatty v0.0.8/go.mod h1:Iq45c/XA43vh69/j3iqttzPXn0bhXyGjM0Hd
 github.com/mattn/go-isatty v0.0.12/go.mod h1:cbi8OIDigv2wuxKPP5vlRcQ1OAZbq2CE4Kysco4FUpU=
```
giautm, line 147: Someone forgot to run `go mod` after running `go generate`. So the content of this file is not correct.

entgql/annotation.go
```diff
@@ -45,6 +45,8 @@ type (
 		Implements []string `json:"Implements,omitempty"`
 		// Directives to add on the field/type.
 		Directives []Directive `json:"Directives,omitempty"`
+		// QueryField allow you expose the Entity on the query as a field.
```
a8m, line 48:
```suggestion
		// QueryField exposes the generated type with the given string under the Query object.
```

Can we also make the name optional, so a call to `QueryField()` will expose the type with generated name (plural form)?

a8m:
Looks great @giautm! Before we merge this, please see my comment.

entgql/annotation.go
```diff
@@ -212,6 +240,24 @@ func (f SkipMode) Is(mode SkipMode) bool {
 	return f&mode != 0
 }
 
+func (c QueryFieldConfig) fieldName(gqlType string) string {
+	if c.Name != "" {
+		return c.Name
+	}
+
```
a8m, line 247:
```suggestion
```

entgql/annotation.go
```diff
@@ -62,6 +64,13 @@ type (
 
 	// SkipMode is a bit flag for the Skip annotation.
 	SkipMode int
+
+	QueryFieldConfig struct {
```
a8m, line 68:
```suggestion
	FieldConfig struct {
```
Allow using it in other types in the future (e.g. Mutation).

PR #290, opened 2022-04-21
entgql: support whereInput on the edge

giautm:
Related issue: https://github.com/ent/ent/issues/2351


entgql/extension.go
```diff
@@ -207,6 +207,8 @@ func (e *Extension) mapScalar(f *gen.Field, op gen.Op) string {
 	switch t := f.Type.Type; {
 	case op.Niladic():
 		return "Boolean"
+	case f.Name == "id":
+		return "ID"
```
a8m, line 211: You can add this to the case below:

```go
case case f.IsEdgeField(), f.Name == "id":
      // ...
```

entgql/internal/todoplugin/ent/schema/user.go
```diff
@@ -35,14 +35,16 @@ func (User) Fields() []ent.Field {
...
+			),
 		field.Float("amount").
 			GoType(Amount(0)).
-			Annotations(entgql.Annotation{
-				Type: "Float",
-			}),
+			Annotations(
+				entgql.Type("Float"),
+				entgql.Skip(entgql.SkipWhereInput),
+			),
```
a8m, line 47: neat 🚀

entgql/internal/todoplugin/entgql.resolvers.go
```diff
@@ -39,7 +39,7 @@ func (r *queryResolver) Nodes(ctx context.Context, ids []int) ([]ent.Noder, erro
 	return r.client.Noders(ctx, ids)
 }
 
-func (r *queryResolver) Todos(ctx context.Context, after *ent.Cursor, first *int, before *ent.Cursor, last *int, orderBy *ent.TodoOrder) (*ent.TodoConnection, error) {
+func (r *queryResolver) Todos(ctx context.Context, after *ent.Cursor, first *int, before *ent.Cursor, last *int, orderBy *ent.TodoOrder, where *ent.TodoWhereInput) (*ent.TodoConnection, error) {
```
a8m, line 42: For the sake of the example, let's use it in the `Paginate` call below.

entgql/internal/todoplugin/gqlgen.yml
```diff
@@ -32,3 +32,12 @@ models:
...
       - entgo.io/contrib/entgql/internal/todouuid/ent/schema/uuidgql.UUID
+  CategoryWhereInput:
+    model:
+      - entgo.io/contrib/entgql/internal/todoplugin/ent.CategoryWhereInput
+  TodoWhereInput:
+    model:
+      - entgo.io/contrib/entgql/internal/todoplugin/ent.TodoWhereInput
+  MasterUserWhereInput:
+    model:
+      - entgo.io/contrib/entgql/internal/todoplugin/ent.MasterUserWhereInput
```
a8m, line 43: Can't you use autobind on the `ent` package like all other examples?

a8m (approved):

entgql/extension.go
giautm: I need to return the "ID" value early

entgql/internal/todoplugin/gqlgen.yml
giautm: On this package, I want to use `@goModel` or map them via the plugin. But, If I add `@goModel` it will break other examples.

PR #293, opened 2022-04-24
entgql: cleanup queryField annotation

giautm:
I change the API to a chain, to be cleaner.
```go
entgql.QueryField().
	Directives(entgql.Deprecated("we don't use this anymore")),
```


a8m (approved):
💯

PR #294, opened 2022-04-24
entgql: adds support for custom predicates on where structs

crossworth:
See [comment](https://github.com/ent/ent/issues/2351#issuecomment-1107518277) for more information.

The only real changes are:
https://github.com/ent/contrib/blob/c0c2a2d3b6fae82472234b30bd981803881076f6/entgql/template/where_input.tmpl#L35

and:
https://github.com/ent/contrib/blob/c0c2a2d3b6fae82472234b30bd981803881076f6/entgql/template/where_input.tmpl#L133

### DIFF
--- a/entgql/internal/todo/generated.go
+++ b/entgql/internal/todo/generated.go
@@ -43,6 +43,7 @@ type Config struct {
 type ResolverRoot interface {
 	Mutation() MutationResolver
 	Query() QueryResolver
+	TodoWhereInput() TodoWhereInputResolver
 }
 
 type DirectiveRoot struct {
@@ -147,6 +148,10 @@ type QueryResolver interface {
 	Todos(ctx context.Context, after *ent.Cursor, first *int, before *ent.Cursor, last *int, orderBy *ent.TodoOrder, where *ent.TodoWhereInput) (*ent.TodoConnection, error)
 	Users(ctx context.Context, after *ent.Cursor, first *int, before *ent.Cursor, last *int, where *ent.UserWhereInput) (*ent.UserConnection, error)
 	Groups(ctx context.Context, after *ent.Cursor, first *int, before *ent.Cursor, last *int, where *ent.GroupWhereInput) (*ent.GroupConnection, error)
+}
+
+type TodoWhereInputResolver interface {
+	CreatedToday(ctx context.Context, obj *ent.TodoWhereInput, data *bool) error
 }
 
 type executableSchema struct {
@@ -714,6 +719,10 @@ enum TodoOrderField {
 input TodoOrder {
   direction: OrderDirection!
   field: TodoOrderField
+}
+
+extend input TodoWhereInput {
+  createdToday: Boolean
 }
 
 type Query {
@@ -6741,6 +6750,17 @@ func (ec *executionContext) unmarshalInputTodoWhereInput(ctx context.Context, ob
 			ctx := graphql.WithPathContext(ctx, graphql.NewPathWithField("hasCategoryWith"))
 			it.HasCategoryWith, err = ec.unmarshalOCategoryWhereInput2ᚕᚖentgoᚗioᚋcontribᚋentgqlᚋinternalᚋtodoᚋentᚐCategoryWhereInputᚄ(ctx, v)
 			if err != nil {
+				return it, err
+			}
+		case "createdToday":
+			var err error
+
+			ctx := graphql.WithPathContext(ctx, graphql.NewPathWithField("createdToday"))
+			data, err := ec.unmarshalOBoolean2ᚖbool(ctx, v)
+			if err != nil {
+				return it, err
+			}
+			if err = ec.resolvers.TodoWhereInput().CreatedToday(ctx, &it, data); err != nil {
 				return it, err
 			}
 		}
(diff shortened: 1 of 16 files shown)


giautm:
The only thing I was considering is, how to support `Or` for custom predicates. We can add more methods to the WhereInput structure to support modifying the predicates, but not expose them as public

crossworth:
@giautm what do you mean by support `Or`? 
About adding methods, we could add a `AddPredicate` and make the predicates slice private, but since they only store the custom predicates I don't see the point.

giautm:
> @giautm what do you mean by support `Or`?

Yep, Because now it always uses the And operator

> About adding methods, we could add an `AddPredicate` and make the predicates slice private, but since they only store the custom predicates I don't see the point.

The API will be more simple, 
```golang
obj.AddPredicates(todo.And(todo.CreatedAtGTE(startOfDay), todo.CreatedAtLTE(endOfDay)))
```

crossworth:
I'm not sure yet about the `Or`, I don't see why it would not work. I mean, the code that [does the or](https://github.com/ent/contrib/blob/c0c2a2d3b6fae82472234b30bd981803881076f6/entgql/internal/todo/ent/gql_where_input.go#L663) is called recursive, when we get to the [`And`](https://github.com/ent/contrib/blob/c0c2a2d3b6fae82472234b30bd981803881076f6/entgql/internal/todo/ent/gql_where_input.go#L868) the or is already applied.
If you could help me understand I would appreciate.

About the `AddPredicates` method, I don't mind adding a method and making the slice private, maybe there is a need to have direct access to the predicates slices, like checking if another predicate is already applied and erroring out if so or removing the predicates if another one is found, but I don't have a user case for it.

a8m:
@giautm, I'm not sure I see a limitation with `Or` predicates. I use the same approach in the `<T>Query` builders in Ent. Do you have any specific examples you think of? 

Regarding the `Predicates` slice, I think it makes sense to keep it public to allow more control over the predicates in case there is more than one resolver involved in the construction of it. However, I don't mind having `AddPredicates` as a method.

a8m (approved):
PR looks fine to me. @crossworth, would you like to update the docs? 
https://entgo.io/docs/tutorial-todo-gql-filter-input

giautm (approved):
LGTM, thank you for your contribution.

entgql/internal/todo/generated.go
```diff
@@ -6743,6 +6752,17 @@ func (ec *executionContext) unmarshalInputTodoWhereInput(ctx context.Context, ob
...
+			var err error
+
+			ctx := graphql.WithPathContext(ctx, graphql.NewPathWithField("createdToday"))
+			data, err := ec.unmarshalOBoolean2ᚖbool(ctx, v)
+			if err != nil {
+				return it, err
+			}
+			if err = ec.resolvers.TodoWhereInput().CreatedToday(ctx, &it, data); err != nil {
+				return it, err
+			}
```
giautm, line 6765: @a8m This PR may affect our custom unmarshal task. Because We don't know this field to unmarshal it. https://github.com/ent/contrib/pull/290

a8m: Agree. Good catch. 

I still think we should not be blocked on this and let @crossworth propose an elegant solution for this.

giautm: It also requires a call to the field resolver, which thing we don't have access to in the Unmarshal context.

crossworth: I don't fully understand the reason why we should have the custom unmarshal for the complete `WhereFilter`, I got the gist from [this message](https://discord.com/channels/885059418646003782/944956624433467492/960812849016471574). That we need to handle the [Field collection](https://entgo.io/docs/tutorial-todo-gql-field-collection), but from what I understand, the field collection handles what we want to return from the resolver and in Ent's case what we want to eager load, I don't see the reason the filtering would be collected or even know in advance, I think I'm missing something...

With that said, I have a question, does this limitation don't apply to the case when using the generation hooks as well?
If I add a custom field (to the graphql part) using `WithSchemaHook` and override the template to add the required field + the filtering condition, I dont see how Ent is supposed to know the type and how to unmarshal it (of course we could require that every custom type have the matching unmarshal/marshal function to delegate to it).

Alternative ideas for allowing easy custom predicates:
- Improve the template adding similar concepts like [`model/fields/additional`](https://entgo.io/docs/schema-fields/#additional-struct-fields), but for the where struct and where filtering, this way we dont have to override the complete template (I'm not sure if this bypass this limitations).
- Create a new entgql annotation to define this "Custom predicates" to the schema and provide the code generation information about the type, this way we wont need to `extend input TodoWhereInput` but would need to think where to provide the predicate function, maybe in the definition itself (I think this would remove this limitations right?)

But we have to be honest with ourselves, both options above are way more complex and hard to the final user, this PR is a two line deal with very little complexity on the Ent side of things, maybe we should try to solve the other side of the equation.

a8m: > but from what I understand, the field collection handles what we want to return from the resolver and in Ent's case what we want to eager load, I don't see the reason the filtering would be collected or even know in advance, I think I'm missing something...

Imagine a query like the follows:
```graphql
query {
  users(limit: 100) {
    edges {
      groups(limit: 10, where: {active: true}) {
        totalCount
      }
    }
  }
}
```
Instead of letting gqlgen resolve the groups of each user (N+1 queries), using eager-loading in the fields-collection stage will resolve all groups of all 100 users will be resolved by entgql in a single query. However, the arguments are not "parsed" in this stage ("users" context) and entgql needs to parse them manually (convert from a raw GraphQL representation to a concrete type). You can see how does the generated resolvers are worked in gqlgen generated files.

PR #304, opened 2022-05-02
entgql: Conditionally include orderBy argument in connection field

scraymondjr:
In working off latest tip I discovered that the connection query fields were always being generated with the `OrderBy` argument referencing an OrderBy type for the node type. This produces invalid graphql schema in the case that no OrderBy type exists for the node type. A quick fix for this is to check if any OrderBy fields are specified and `SkipOrderField` is not set for the type and pass the result of this check to `ConnectionField` method. This same check needs to be applied at both usages of `ConnectionField` - in building the schema for a node and building the edges for the node.

### DIFF
--- a/entgql/schema.go
+++ b/entgql/schema.go
@@ -240,13 +240,16 @@ func (e *schemaGenerator) buildTypes(s *ast.Schema) error {
 			names := paginationNames(gqlType)
 
 			s.AddTypes(names.TypeDefs()...)
+
+			var hasOrderBy bool
 			if len(enumOrderByValues) > 0 && !ant.Skip.Is(SkipOrderField) {
+				hasOrderBy = true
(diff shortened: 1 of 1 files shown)


scraymondjr:
Creating this PR before writing tests to check if this change is valid. Would appreciate a sanity check that this issue does in fact need to be addressed and then will happily add test coverage to complete the PR.

giautm:
Hey, thank @scraymondjr for this PR. I will include this change in https://github.com/ent/contrib/pull/305, because we need to update the Go template too.

giautm:
Hello @scraymondjr, can you try this branch to verify your issue? https://github.com/ent/contrib/pull/305

scraymondjr:
Hey @giautm I tried with the latest commit in contrib and it looks like this is fixed! Thank you.

PR #306, opened 2022-05-04
entgql: update examples to using schema generator

giautm:


entgql/extension.go
```diff
@@ -215,7 +225,7 @@ func (e *Extension) genSchemaHook() gen.Hook {
 // in the template list and returns its index.
 func (e *Extension) whereExists() (int, bool) {
 	for i := range e.templates {
-		if e.templates[i] == WhereTemplate {
+		if e.templates[i].Name() == "gql_where_input" {
```
a8m, line 228:
```suggestion
		if e.templates[i].Name() == WhereTemplate.Templates()[1].Name() {
```

Use the parsed one instead of using a raw string here. The result will be the same.

entgql/internal/todo/ent.resolvers.go
```diff
@@ -17,7 +17,47 @@ package todo
 // This file will be automatically regenerated based on the schema, any resolver implementations
 // will be copied through when generating and any unknown code will be moved to the end.
 
+import (
```
a8m, line 20: Isn't this file a duplication of todo.resolvers.go? It's because the configuration in gqlgen is "follow-schema". Either rename the generated GrpahQL schema to "todo.graphql" or delete https://github.com/ent/contrib/blob/master/entgql/internal/todo/todo.resolvers.go.

entgql/internal/todo/ent/entc.go
```diff
@@ -17,14 +17,15 @@ import (
...
+		entgql.WithConfigPath("./gqlgen.yml"),
+		entgql.WithSchemaGenerator(),
+		entgql.WithSchemaPath("./ent.graphql"),
+		entgql.WithWhereInputs(true),
 	)
 	if err != nil {
 		log.Fatalf("creating entgql extension: %v", err)
 	}
-	err = entc.Generate("./schema", &gen.Config{
+	err = entc.Generate("./ent/schema", &gen.Config{
```
a8m, line 28: Please add a comment explaining the codegen is executed from internal/todo and not internal/todo/ent.

a8m (commented):
Looks good, but see comments, please.

Also, I think it's time to update both the docs and https://github.com/a8m/ent-graphql-example.

entgql/internal/todo/ent.resolvers.go
giautm: The todo.resolvers.go still have resolvers for todo's mutation. So, I don't think I have to change anything.

https://github.com/ent/contrib/blob/giautm/update-todo-example/entgql/internal/todo/todo.resolvers.go

entgql/extension.go
```diff
@@ -215,7 +225,7 @@ func (e *Extension) genSchemaHook() gen.Hook {
 // in the template list and returns its index.
 func (e *Extension) whereExists() (int, bool) {
 	for i := range e.templates {
-		if e.templates[i] == WhereTemplate {
+		if e.templates[i].Name() == WhereTemplate.Templates()[1].Name() {
```
giautm, line 228: @a8m why the index is 1, and not is 0?

a8m (approved):

PR #310, opened 2022-05-07
entgql: extend docs for Bind and Unbind annotation options

dorlib:


entgql/annotation.go
```diff
@@ -123,22 +123,46 @@ func OrderField(name string) Annotation {
...
 
-// Bind returns a binding annotation.
+//Bind returns a binding annotation.
+// entgql.Bind() main purpose's in the Ent schema definition is to improve the efficiency of queries using GraphQL.
+// The Bind() annotation is enabled by default, use Unbind() to opt-out.
 //
-// No-op function to avoid breaking the existing schema.
-// You can safely remove this function from your scheme.
-//
-// Deprecated: the Bind option predates the Unbind option, and it is planned
```
giautm, line 131: @dorlib hello, this annotation was planned to remove. So, why did you remove our deprecated note?

dorlib: The purpose was to extand the docs about this option untill then, and also in order to understand Unbind better.
I think i've added this note before the example and i would like to hear improvement suggestions to this commit 🙂

a8m (commented):
Thanks @dorlib 

I'll suggest a change to the Bind doc later.

entgql/annotation.go
giautm: Please keep the keyword `Deprecated:` as is. To allow go linter mask the function is deprecated in their code.

entgql/annotation.go
```diff
@@ -123,22 +123,46 @@ func OrderField(name string) Annotation {
...
+// the Bind option predates the Unbind option, and it is planned
```
giautm, line 130:
```suggestion
// Deprecated: The Bind option predates the Unbind option, and it is planned
```

a8m (approved):
Excellent @dorlib 

Thanks 🙏

PR #315, opened 2022-05-12
all: upgrade ent to latest

giautm:


entproto/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM golang:1.17.3
+FROM golang:1.18.2
```
giautm, line 1: @rotemtam: I need to do some changes in dockerfile/README to able run go generate, can you review it?

a8m (approved):

PR #324, opened 2022-05-25
entgql: fixed using wrong Annotation from the Node

giautm:


entgql/schema.go
```diff
@@ -178,6 +182,11 @@ func (e *schemaGenerator) buildTypes(g *gen.Graph, s *ast.Schema) error {
 						return err
 					}
 					if def != nil {
+						if s.Types[def.Name] != nil {
+							log.Println(fmt.Sprintf("entgql: the %s.%s field uses the previously defined GQL type: %s",
+								node.Name, f.Name, def.Name))
```
a8m, line 187: Why this logging is needed? I try to follow the unix way and avoid printing logs in non-verbose mode. It also feels like a common thing and I wouldn't want to see it in every "go generate" execution.

a8m (commented):
Looks good, but see the comment.

entgql/schema.go
giautm: @a8m: This log is a warning for users in case they have a mistake when having two Enum with the same default name. For example `Todo.status` and `Category.status` will generate same enum `Status`. Possible fix is using the `entgql.Type()` annotation in one of them.

a8m: I think since GraphQL has a global schema, the generated enum for `Todo.status` should be `TodoStatus` and not `Status`. Also, I don't like the idea of printing in codegen time. I prefer to either fail or ignore this if "it's working". Ent today does not print in codegen and I think also in ent/contrib, so I'd avoid adding this new behavior.

giautm:
I remove the log and rename the default enum name in the commit https://github.com/ent/contrib/pull/324/commits/67a8aa712fc16a610ded495a634cf1530cce5915. To use the old name to avoid BC on your graph, you can use `entgql.Type("OldName")` on the enum field.

entgql/schema.go
```diff
@@ -651,6 +661,11 @@ func (e *schemaGenerator) mapScalar(f *gen.Field, ant *Annotation, typeFilter fu
 		} else {
 			scalar = scalar[strings.LastIndexByte(scalar, '.')+1:]
 		}
+		if f.IsEnum() {
+			// Use the GQL type for enum prefix
+			// the Todo.status field will have the enum with name: TodoStatus
+			scalar = gqlType + scalar
```
a8m, line 667:
```suggestion
			// Use the GQL type as enum prefix. e.g. Todo.status
			// will generate an enum named "TodoStatus".
			scalar = gqlType + scalar
```

a8m (approved):
💯

PR #325, opened 2022-05-27
entgql: fixed edge schema compatible

giautm:
This PR is not fully supported edge schema, it's just fixing the compatibility. So, you can use the edge schema with the EntGQL extension.

Some ideas:
- If the edge schema uses a non-composite id key, it still can expose as a normal edge.
- The edge schema can't expose on the graph if it's using the composite primary key.
- We may support generating more fields in the WhereInput from the edge schema.

https://github.com/ent/ent/pull/2560


entgql/template.go
```diff
@@ -304,6 +304,9 @@ func mutationInputs(nodes []*gen.Type) ([]*MutationDescriptor, error) {
 func filterNodes(nodes []*gen.Type, skip SkipMode) ([]*gen.Type, error) {
 	filteredNodes := make([]*gen.Type, 0, len(nodes))
 	for _, n := range nodes {
+		if n.IsEdgeSchema() {
```
a8m, line 307: We want to skip edge schemas with composite primary keys, right? Not all edge schemas.

a8m: Cool.

a8m (approved):

a8m:
Thanks @giautm 🙏

PR #328, opened 2022-06-04
entgql: Allow using Relay Connection without Node support

tadeokondrak:
Motivation: I want to use the Relay Connection support, but without exposing the Node interface, as the Node interface doesn't seem to use field resolvers and instead reads data directly.

### DIFF
--- a/entgql/extension.go
+++ b/entgql/extension.go
@@ -134,10 +134,27 @@ func WithWhereInputs(b bool) ExtensionOption {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


a8m:
> as the Node interface doesn't seem to use field resolvers and instead reads data directly.

Can you elaborate on what you mean by this?

tadeokondrak:
Tested it out - was looking at the wrong Node implementations, the Node interface does use field resolvers, so this isn't necessary.

PR #329, opened 2022-06-05
feat(entoas): added field annotation for skip flag

badygin-yakov:
Added Ignored field to Annotation struct to support ignored filed in spec (useful for deletedAt field or system fields don't need in spec).


a8m:
cc @masseelch.

Also, please use Skip instead of Ignored to keep it consistent with the rest of the packages here (e.g. entgql).

badygin-yakov:
@a8m I can rename field or need realize logic [SkipMode](https://github.com/ent/contrib/blob/78aa88292a6251eaaae6d5c79dd41a7124c41006/entgql/annotation.go#L83-L107)?

badygin-yakov:
@masseelch review pls

masseelch:
@badygin-yakov I am not sure what this PR does. For what I can see it adds an annotation but does not do anything with it. Can you explain?

entoas/generator.go
```diff
@@ -541,6 +548,11 @@ func OgenSchema(f *gen.Field) (*ogen.Schema, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if ant.Skip {
+		return nil, nil
+	}
+
```
masseelch, line 555: I prefer to have this done more explicit by either removing the field [here](https://github.com/ent/contrib/blob/e738ef866e51c76d31fcc71919a7986d2e403719/entoas/generator.go#L59) or maybe [here](https://github.com/ent/contrib/blob/e738ef866e51c76d31fcc71919a7986d2e403719/entoas/generator.go#L127)

masseelch (changes requested):
Hey @badygin-yakov, thanks for this PR.

Please see my comment.

entoas/generator.go
badygin-yakov: Okey, I refactored code and made skip logic more explicit

entoas/generator.go
```diff
@@ -126,6 +126,15 @@ func schemas(g *gen.Graph, spec *ogen.Spec) error {
 // addSchemaFields adds the given gen.Field slice to the ogen.Schema.
 func addSchemaFields(s *ogen.Schema, fs []*gen.Field) error {
 	for _, f := range fs {
+		ant, err := FieldAnnotation(f)
+		if err != nil {
+			return err
+		}
+
```
masseelch, line 133:
```suggestion
```

entoas/generator.go
```diff
@@ -126,6 +126,15 @@ func schemas(g *gen.Graph, spec *ogen.Spec) error {
...
+		if ant.Skip {
+			continue
+		}
+
```
masseelch, line 137:
```suggestion
```

entoas/generator.go
```diff
@@ -678,7 +687,12 @@ func reqBody(n *gen.Type, op Operation) (*ogen.RequestBody, error) {
 		if err != nil {
 			return nil, err
 		}
-		if (a != nil && !a.ReadOnly) && (op == OpCreate || !f.Immutable) {
+
```
masseelch, line 690:
```suggestion
```

entoas/generator.go
```diff
@@ -678,7 +687,12 @@ func reqBody(n *gen.Type, op Operation) (*ogen.RequestBody, error) {
...
+		if a.ReadOnly || a.Skip {
+			continue
+		}
+
```
masseelch, line 694:
```suggestion
```

entoas/generator.go
```diff
@@ -678,7 +687,12 @@ func reqBody(n *gen.Type, op Operation) (*ogen.RequestBody, error) {
 		if err != nil {
 			return nil, err
 		}
-		if (a != nil && !a.ReadOnly) && (op == OpCreate || !f.Immutable) {
+
+		if a.ReadOnly || a.Skip {
```
masseelch, line 691: The check for `a != nil` is required, isn't it?

entoas/internal/simple/schema/category.go
```diff
@@ -15,10 +15,11 @@
 package schema
 
 import (
-	"entgo.io/contrib/entoas"
 	"entgo.io/ent"
 	"entgo.io/ent/schema/edge"
 	"entgo.io/ent/schema/field"
+
```
masseelch, line 21:
```suggestion
```

masseelch (changes requested):
Nice! See some suggestions, then I can merge :-)

entoas/generator.go
badygin-yakov: Why need check? `a` can be nil only if `FieldAnnotation` return err. But this case we handle on [684-687](https://github.com/ent/contrib/pull/329/files#diff-8a343f04c5df9bc0b27ce52ca7989745e813d4779329f49f69ca18d74963c993R684-R687)

masseelch (approved):

masseelch:
Please fix the CI, then I will merge

PR #339, opened 2022-06-20
URGENT, please: add constructor for order fields

frederikhors:
This is needed because where are converting internal struct to ent ones.


a8m:
Hey @frederikhors and thanks for the contribution. Please, use a [custom template](https://entgo.io/docs/templates/) to expose this API for your use case, and let me know if you need help with it. I prefer to keep these struct fields private and not expose such an API because we plan to improve the ordering capability of entgql soon and this will introduce breaking changes (e.g. multiple fields support).

Closing, and thanks again.

PR #340, opened 2022-06-21
use uint64 for TotalCount

frederikhors:
I think using uint instead of int because the number can never be negative.


masseelch:
Hey @frederikhors, please re-run the code generator for this PR :-)

frederikhors:
Do you mean `go generate ./...`?

frederikhors:
I can do this, but first tell me if you want to use `uint` instead of `int` so I can finish the change wherever needed.

frederikhors:
Ok. Should we leave this open to track in future major version?

a8m:
> Ok. Should we leave this open to track in future major version?

We ~promise backward compatibility for Ent even for new major versions. My idea is to provide something like [`facebook/codemod`](https://github.com/facebookarchive/codemod) but for Go. This will give us the freedom to introduce gradual changes for Ent API without affecting the users that much. 

Closing as it's already on my list.

PR #344, opened 2022-06-27
entgql: fixed edge filter

giautm:
Thanks @amrnt for reporting and provide the test cases, ❤️

### DIFF
--- a/entgql/internal/todo/todo_test.go
+++ b/entgql/internal/todo/todo_test.go
@@ -640,6 +640,118 @@ func (s *todoTestSuite) TestPaginationFiltering() {
 	})
 }
 
+func (s *todoTestSuite) TestEdgesFiltering() {
+	ctx := context.Background()
+	tr := s.ent.Todo.Create().
+		SetText("t1").
+		SetStatus(todo.StatusInProgress).
+		SaveX(ctx)
+
+	t1 := s.ent.Todo.Create().
+		SetText("t2 0").
+		SetStatus(todo.StatusInProgress).
+		SetParent(tr).
+		SaveX(ctx)
+	s.ent.Todo.Create().
+		SetText("t2 1").
+		SetStatus(todo.StatusCompleted).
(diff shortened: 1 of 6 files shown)


amrnt (approved):
Looking good 💯 🥇

entgql/internal/todo/todo_test.go
```diff
@@ -640,6 +640,118 @@ func (s *todoTestSuite) TestPaginationFiltering() {
 	})
 }
 
+func (s *todoTestSuite) TestEdgesFiltering() {
```
a8m, line 643: I think we can move this test case to the [`NestedPagination`/`NestedConnection`](https://github.com/ent/contrib/blob/master/entgql/internal/todo/todo_test.go#L1266) one.

Also, in this test case, you don't really test the `Query<E>` method, because the edges are resolved/loaded by the fields-collector (using eager-loading). 

@amrnt, how do you use this API? Do you call these methods directly?

entgql/internal/todo/todo_test.go
```diff
@@ -640,6 +640,118 @@ func (s *todoTestSuite) TestPaginationFiltering() {
...
+	query := `query todos($id: ID!, $lv2Status: TodoStatus!) {
+		todos(where:{id: $id}) {
+			edges {
+				node {
+					childrenLevel0: children(where: {statusNEQ: COMPLETED}) {
+						totalCount
+						edges {
+							node {
+								id
+								childrenLevel1: children(where: {statusNEQ: $lv2Status}) {
```
a8m, line 681: OK, I see now. It's called `childrenLevel1` and that's why it's not loaded. 

I still want to give it a look before merging it.

entgql/internal/todo/todo_test.go
a8m: OK, I see now. It's called `childrenLevel1` and that's why it's not loaded. 

I still want to give it a look before merging it.

PR #358, opened 2022-07-21
entgql: changed receiver to `q` to avoid conflict

giautm:


giautm:
Close, let cleanup later with the update in ent.

PR #361, opened 2022-08-04
bumped pkg versions

ankitpatial:
I am getting an error "Unexpected package creation during export data loading" 
on comand "ent generate" (after moving to go1.19) and this  seems linked to the contrib package.

Might be its happening because of some old package, So i bumped version for all packages and trand go test ./... command to see if all tests runs ok.


ankitpatial:
Please ignore it. I don't think my issue will be resolved with just bumping version. I am trying to dig in the issue

PR #375, opened 2022-09-02
entgql: Use nil to check that a generated enum satisfies interface 

Adam-Mustafa:
Fixes https://github.com/ent/ent/issues/2517

Changes the interface check from trying to instantiate a string enum to checking the nil pointer type.


giautm (approved):

a8m (approved):
Thanks for the contribution, @Adam-Mustafa 🙏🏻

PR #380, opened 2022-09-20
feat(entgql): support child value list for directive annotation

calmonr:
Hi 👋 That's my first time contributing to `Ent` and I must say I'm enjoying it so far.

This PR allows the creation of a directive annotation using a child value list, for example:

```
directive @hasPermissions(permissions: [String!]!) on OBJECT | FIELD_DEFINITION
```

The original implementation only allowed the user to specify a `string` as a value:

https://github.com/ent/contrib/blob/462c4e5698731377eb38f5a61b16518bb7295341/entgql/annotation.go#L61-L65

https://github.com/ent/contrib/blob/462c4e5698731377eb38f5a61b16518bb7295341/entgql/annotation.go#L518-L528

Now you can specify a child value list:

```golang
func HasPermissions(permissions []string) entgql.Directive {
	return entgql.NewDirective("hasPermissions", entgql.DirectiveArgument{
		Name:     "permissions",
		Kind:     ast.ListValue,
		Children: entgql.DirectiveChildren(permissions),
	})
}
```

The `entgql.DirectiveChildren` is a function that coverts a slice of `string` to `ast.ChildValueList`.

You can use it like this:

```golang
// Annotations returns Group annotations.
func (Group) Annotations() []schema.Annotation {
	return []schema.Annotation{
		entgql.RelayConnection(),
		entgql.QueryField(),
		entgql.Directives(
			annotation.HasPermissions([]string{"ADMIN", "MODERATOR"}),
		),
	}
}
```

The result will be on your `ent.graphql`:

```graphql
type Group implements Node @hasPermissions(permissions: ["ADMIN","MODERATOR"]) {
```

I'm also accepting suggestions for the `entgql.DirectiveChildren` function name if you don't like it.

Thank you.

### DIFF
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -59,9 +59,10 @@ type (
 
 	// DirectiveArgument return a GraphQL directive argument
 	DirectiveArgument struct {
-		Name  string        `json:"name,omitempty"`
-		Value string        `json:"value,omitempty"`
-		Kind  ast.ValueKind `json:"kind,omitempty"`
+		Name     string             `json:"name,omitempty"`
+		Children ast.ChildValueList `json:"children,omitempty"`
+		Value    string             `json:"value,omitempty"`
+		Kind     ast.ValueKind      `json:"kind,omitempty"`
 	}
 
 	// SkipMode is a bit flag for the Skip annotation.
@@ -512,6 +513,23 @@ func NewDirective(name string, args ...DirectiveArgument) Directive {
 		Name:      name,
 		Arguments: args,
 	}
+}
+
+func DirectiveChildren(s []string) ast.ChildValueList {
+	var list ast.ChildValueList
+
+	for _, v := range s {
+		child := &ast.ChildValue{
+			Value: &ast.Value{
+				Raw:  v,
+				Kind: ast.StringValue,
+			},
+		}
+
+		list = append(list, child)
+	}
+
+	return list
 }
 
 // Deprecated create `@deprecated` directive to apply on the field/type
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -53,16 +53,8 @@ type (
 
 	// Directive to apply on the field/type.
 	Directive struct {
-		Name      string              `json:"name,omitempty"`
-		Arguments []DirectiveArgument `json:"arguments,omitempty"`
-	}
-
-	// DirectiveArgument return a GraphQL directive argument
-	DirectiveArgument struct {
-		Name     string             `json:"name,omitempty"`
-		Children ast.ChildValueList `json:"children,omitempty"`
-		Value    string             `json:"value,omitempty"`
-		Kind     ast.ValueKind      `json:"kind,omitempty"`
+		Name      string          `json:"name,omitempty"`
+		Arguments []*ast.Argument `json:"arguments,omitempty"`
 	}
 
 	// SkipMode is a bit flag for the Skip annotation.
@@ -508,39 +500,23 @@ var (
 
 // NewDirective returns a GraphQL directive
 // to use with the entgql.Directives annotation.
-func NewDirective(name string, args ...DirectiveArgument) Directive {
+func NewDirective(name string, args ...*ast.Argument) Directive {
 	return Directive{
 		Name:      name,
 		Arguments: args,
 	}
 }
 
-// DirectiveListArgument returns an argument with the list of values.
-func DirectiveListArgument(name string, kind ast.ValueKind, rawValues ...string) DirectiveArgument {
-	list := make(ast.ChildValueList, len(rawValues))
-	for i, v := range rawValues {
-		list[i] = &ast.ChildValue{
-			Value: &ast.Value{
-				Raw:  v,
-				Kind: kind,
-			},
-		}
-	}
-	return DirectiveArgument{
-		Children: list,
-		Kind:     ast.ListValue,
-		Name:     name,
-	}
-}
-
 // Deprecated create `@deprecated` directive to apply on the field/type
 func Deprecated(reason string) Directive {
-	var args []DirectiveArgument
+	var args []*ast.Argument
 	if reason != "" {
-		args = append(args, DirectiveArgument{
-			Name:  "reason",
-			Kind:  ast.StringValue,
-			Value: reason,
+		args = append(args, &ast.Argument{
+			Name: "reason",
+			Value: &ast.Value{
+				Raw:  reason,
+				Kind: ast.StringValue,
+			},
 		})
 	}
 	return NewDirective("deprecated", args...)
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -515,22 +515,23 @@ func NewDirective(name string, args ...DirectiveArgument) Directive {
 	}
 }
 
-// DirectiveChildren returns a ChildValueList for the directive.
-func DirectiveChildren(k ast.ValueKind, s []string) ast.ChildValueList {
-	var list ast.ChildValueList
 
-	for _, v := range s {
-		child := &ast.ChildValue{
+// DirectiveListArgument returns an argument with the list of values.
+func DirectiveListArgument(name string, kind ast.ValueKind, rawValues ...string) DirectiveArgument {
+	list := make(ast.ChildValueList, len(rawValues))
+	for i, v := range rawValues {
+		list[i] = &ast.ChildValue{
 			Value: &ast.Value{
 				Raw:  v,
-				Kind: k,
+				Kind: kind,
 			},
 		}
-
-		list = append(list, child)
 	}
-
-	return list
+	return DirectiveArgument{
+		Children: list,
+		Kind:     ast.ListValue,
+		Name:     name,
+	}
 }
 
 // Deprecated create `@deprecated` directive to apply on the field/type
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -515,7 +515,6 @@ func NewDirective(name string, args ...DirectiveArgument) Directive {
 	}
 }
 
-
 // DirectiveListArgument returns an argument with the list of values.
 func DirectiveListArgument(name string, kind ast.ValueKind, rawValues ...string) DirectiveArgument {
 	list := make(ast.ChildValueList, len(rawValues))
--- a/entgql/annotation.go
+++ b/entgql/annotation.go
@@ -515,14 +515,15 @@ func NewDirective(name string, args ...DirectiveArgument) Directive {
 	}
 }
 
-func DirectiveChildren(s []string) ast.ChildValueList {
+// DirectiveChildren returns a ChildValueList for the directive.
+func DirectiveChildren(k ast.ValueKind, s []string) ast.ChildValueList {
 	var list ast.ChildValueList
 
 	for _, v := range s {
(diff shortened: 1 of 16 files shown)


calmonr:
I just talked with @crossworth and he suggested that `entgql.DirectiveChildren` accepts the kind as an argument to allow cases like `entgql.DirectiveChildren(ast.IntValue, []string{"10", "20", "30")`.

Thank you for the suggestion.

entgql/annotation.go
```diff
@@ -59,9 +59,10 @@ type (
 
 	// DirectiveArgument return a GraphQL directive argument
 	DirectiveArgument struct {
-		Name  string        `json:"name,omitempty"`
-		Value string        `json:"value,omitempty"`
-		Kind  ast.ValueKind `json:"kind,omitempty"`
+		Name     string             `json:"name,omitempty"`
+		Children ast.ChildValueList `json:"children,omitempty"`
+		Value    string             `json:"value,omitempty"`
+		Kind     ast.ValueKind      `json:"kind,omitempty"`
```
a8m, line 65: Maybe we can squash these 3  into an `ast.Value` and compose this object on the option.

entgql/annotation.go
```diff
@@ -514,6 +515,24 @@ func NewDirective(name string, args ...DirectiveArgument) Directive {
 	}
 }
 
+// DirectiveChildren returns a ChildValueList for the directive.
+func DirectiveChildren(k ast.ValueKind, s []string) ast.ChildValueList {
```
a8m, line 519: I'd say, accepting an `ast.Value` feels more flexible to me.

entgql/internal/todo/resolver.go
```diff
@@ -26,5 +26,8 @@ type Resolver struct{ client *ent.Client }
 func NewSchema(client *ent.Client) graphql.ExecutableSchema {
 	return NewExecutableSchema(Config{
 		Resolvers: &Resolver{client},
+		Directives: DirectiveRoot{
+			HasPermissions: HasPermission(),
+		},
```
a8m, line 31: I think it's preferred to keep our examples simpler. The expected testing from our side is to check the generated schema, the rest if gqlgen logic.

entgql/internal/todo/todo.graphql
```diff
@@ -1,3 +1,5 @@
+directive @hasPermissions(permissions: [String!]!) on OBJECT | FIELD_DEFINITION
```
a8m, line 1: 💯

entgql/schema_test.go
```diff
@@ -96,7 +96,7 @@ type Friendship {
   user: User!
   friend: User!
 }
-type Group {
+type Group @hasPermissions(permissions: ["ADMIN","MODERATOR"]) {
```
a8m, line 99: 💯

entgql/annotation.go
calmonr: I honestly think it makes sense, will make it more flexible. In terms of changes, do you mean that instead of duplicating each field from `ast.Value` on `DirectiveArgument` use an actual `ast.Value` as a field or remove the `DirectiveArgument` and use `ast.Value` altogether?

entgql/internal/todo/resolver.go
calmonr: I completely understand. I put this here because I can't run the tests without it, cause of the unimplemented directive.
> [{"message":"directive hasPermissions is not implemented","path":["users","edges",0,"node","groups","edges",0,"node"]}]

entgql/annotation.go
calmonr: The idea for `DirectiveChildren` is to be used as a complement to `DirectiveArgument` in the `Children` field. I'm a little bit confused about how to accept `ast.Value` here. Perhaps if all the changes discussed above using the actual `ast.Value` instead of `DirectiveArgument` or something like that we don't need the `DirectiveChildren` function anymore as its idea is just to convert from a slice of strings to a `ast.ChildValueList` to be used in the `Children` of `ast.Value`.

entgql/annotation.go
```diff
@@ -514,6 +515,24 @@ func NewDirective(name string, args ...DirectiveArgument) Directive {
...
+				Raw:  v,
+				Kind: k,
+			},
+		}
+
+		list = append(list, child)
+	}
+
+	return list
+}
```
giautm, line 534:
```suggestion

// DirectiveListArgument returns an argument with the list of values.
func DirectiveListArgument(name string, kind ast.ValueKind, rawValues ...string) DirectiveArgument {
	list := make(ast.ChildValueList, len(rawValues))
	for i, v := range rawValues {
		list[i] = &ast.ChildValue{
			Value: &ast.Value{
				Raw:  v,
				Kind: kind,
			},
		}
	}
	return DirectiveArgument{
		Children: list,
		Kind:     ast.ListValue,
		Name:     name,
	}
}
```

calmonr: I really liked this suggestion as it can handle cases of default values for directives (I imagine). I remember discussing with @crossworth about that. He mentioned that that would be cases where directives could have default values and I suggested using variadic functions to handle it. I think it does the job really well. Thank you for that.

What do you think @crossworth?

calmonr: If you guys are okay with the suggestion (I'm) I can accept (commit) and push another one to fix the parts that are using the old version of `DirectiveChildren` to use `DirectiveListArgument`. Just let me know.

calmonr: The end result:

```golang
func HasPermissions(permissions []string) entgql.Directive {
	return entgql.NewDirective(
		"hasPermissions",
		entgql.DirectiveListArgument("permissions", ast.StringValue, permissions...),
	)
}
```

I included the tests as well for the function.

giautm (approved):
LGTM, wdyt @a8m ?

entgql/annotation.go
a8m: Yes, thanks.

giautm: @a8m `DirectiveArgument` was renamed to `DirectiveListArgument` as my suggestion. `DirectiveListArgument` is a helper that allows ppl build arguments quickly and with less code. If they want to customize more, just build the Children manually and pass it to the directive argument.

calmonr: Here's an usage example for the @giautm suggestion: https://github.com/ent/contrib/pull/380#discussion_r977438993

giautm: > Yes, thanks.

I got your idea here. But, it's a breaking change. Do we need to break? if yes, we can use `ast.Argument` to replace `entgql.DirectiveArgument `

a8m:
> LGTM, wdyt @a8m ?

@giautm, see my comment above about using `ast.Value`.

I'm not sure why we need to create custom wrappers to the `ast` package, instead of accepting `ast.Value` and maybe adding the fluent API to `gqlparser/ast`?  

```go
Annotations(
	entgql.Directives(
		entgql.Deprecated("use category instead"),
		entgql.NewDirective("hasPermission", entgql.DirectiveArgument{
			Name: "permissions",
			Value: &ast.Value{
				Kind: ast.ListValue,
				Children: ast.ChildValueList{
					&ast.ChildValue{Value: &ast.Value{Raw: "bar", Kind: ast.StringValue}},
					&ast.ChildValue{Value: &ast.Value{Raw: "baz", Kind: ast.StringValue}},
				},
			},
		}),
	),
),
```

giautm:
Another idea is replace entgql.DirectiveArgument with ast.Argument, but just keep it as and alias type.

```go
type DirectiveArgument = ast.Argument
```

then
```go
// NewDirective returns a GraphQL directive
// to use with the entgql.Directives annotation.
func NewDirective(name string, args ...ast.Argument) Directive {
	return Directive{
		Name:      name,
		Arguments: args,
	}
}
```

a8m:
Discussed with @giautm, and he suggested defining `type DirectiveArgument = ast.Argument`. I like it.

calmonr:
Do you mean like this? If you guys agree with this I can push the commit.

```golang
DirectiveArgument = ast.Argument
```

```golang
// NewDirective returns a GraphQL directive
// to use with the entgql.Directives annotation.
func NewDirective(name string, args ...DirectiveArgument) Directive {
	return Directive{
		Name:      name,
		Arguments: args,
	}
}
```

```golang
func (e *schemaGenerator) buildDirectives(directives []Directive) ast.DirectiveList {
	list := make(ast.DirectiveList, 0, len(directives))
	for _, d := range directives {
		args := make(ast.ArgumentList, 0, len(d.Arguments))
		for _, a := range d.Arguments {
			args = append(args, &a)
		}
		list = append(list, &ast.Directive{
			Name:      d.Name,
			Arguments: args,
		})
	}
	return list
}
```

```golang
// DirectiveListArgument returns an argument with the list of values.
func DirectiveListArgument(name string, kind ast.ValueKind, rawValues ...string) DirectiveArgument {
	list := make(ast.ChildValueList, len(rawValues))
	for i, v := range rawValues {
		list[i] = &ast.ChildValue{
			Value: &ast.Value{
				Raw:  v,
				Kind: kind,
			},
		}
	}
	return DirectiveArgument{
		Name: name,
		Value: &ast.Value{
			Kind:     ast.ListValue,
			Children: list,
		},
	}
}
```

```golang
// Deprecated create `@deprecated` directive to apply on the field/type
func Deprecated(reason string) Directive {
	var args []DirectiveArgument
	if reason != "" {
		args = append(args, DirectiveArgument{
			Name: "reason",
			Value: &ast.Value{
				Raw:  reason,
				Kind: ast.StringValue,
			},
		})
	}
	return NewDirective("deprecated", args...)
}
```

```golang
func TestDirectiveListArgument(t *testing.T) {
	t.Parallel()
	expected := entgql.DirectiveArgument{
		Name: "foo",
		Value: &ast.Value{
			Kind: ast.ListValue,
			Children: ast.ChildValueList{
				&ast.ChildValue{Value: &ast.Value{Raw: "bar", Kind: ast.StringValue}},
				&ast.ChildValue{Value: &ast.Value{Raw: "baz", Kind: ast.StringValue}},
			},
		},
	}
	arg := entgql.DirectiveListArgument("foo", ast.StringValue, "bar", "baz")
	require.Equal(t, expected.Name, arg.Name)
	require.Equal(t, expected.Value.Kind, arg.Value.Kind)
	require.Equal(t, expected.Value.Children, arg.Value.Children)
}
```

calmonr:
Just committed the changes.

Replaced `entgql.DirectiveArgument` in favor of `ast.Argument` and removed the helper `entgql.DirectiveListArgument` that doesn't make much sense now.

entgql/schema.go
```diff
@@ -361,13 +361,7 @@ func (e *schemaGenerator) buildDirectives(directives []Directive) ast.DirectiveL
 	for _, d := range directives {
 		args := make(ast.ArgumentList, 0, len(d.Arguments))
 		for _, a := range d.Arguments {
```
giautm, line 363: This loop is not need anymore. Just assign from the d.Argument

giautm (approved):
Please update the last comment, ;)

entgql/schema.go
calmonr: I totally missed that one, thank you for pointing it out. It's refactored 👍🏻 .

a8m (approved):
Thanks for the contribution, @calmonr.

PR #385, opened 2022-10-04
entgql: accept non-gql context in generated edge methods

a8m:
In case the context is non-gql, we fllback to the regular eager-loading flow (i.e. With<E>)


giautm (approved):

a8m:
Thanks for the review, @giautm <3

PR #387, opened 2022-10-05
entgql: create field Map example

giautm:


entgql/internal/todo/ent.resolvers.go
```diff
@@ -53,11 +54,43 @@ func (r *queryResolver) Users(ctx context.Context, after *ent.Cursor, first *int
 		)
 }
 
+func (r *todoResolver) Init(ctx context.Context, obj *ent.Todo) (map[string]interface{}, error) {
```
a8m, line 57: You can add `Map` [here](https://github.com/ent/contrib/blob/master/entgql/internal/todo/gqlgen.yml#L23) to avoid these mappers: 
```diff
models:
  ID:
    model:
      - github.com/99designs/gqlgen/graphql.IntID
  Uint64:
    model:
      - github.com/99designs/gqlgen/graphql.Uint64
+  Map:
+    model:
+      - github.com/99designs/gqlgen/graphql.Map
```

entgql/internal/todo/ent/schema/todo.go
```diff
@@ -29,6 +29,8 @@ type Todo struct {
 	ent.Schema
 }
 
+type Map map[string]any
```
a8m, line 32: This type is not needed IMO:

```go
field.JSON("init", map[string]any{}).
	Optional().
	Annotations(entgql.Type("Map")),
```

giautm: Screen Shot 2022-10-06 at 13 40 26
gqlgen generate wrong type if we don't create a new type for the map.

a8m (approved):
💯 

Great thing, @giautm.

PR #390, opened 2022-10-10
feat(entproto): add support for custom package in `gen-entgrpc`

marcozac:
Add `go_package` flag and `$entPackage` var in templates

### DIFF
--- a/entproto/cmd/protoc-gen-entgrpc/main.go
+++ b/entproto/cmd/protoc-gen-entgrpc/main.go
@@ -31,6 +31,7 @@ import (
 
 var (
 	entSchemaPath *string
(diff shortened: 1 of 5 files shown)


rotemtam:
Hey @marcozac !

Thanks for the contribution. Could you please elaborate more on this change, why/when is it needed?

marcozac:
> Hey @marcozac !
> 
> Thanks for the contribution. Could you please elaborate more on this change, why/when is it needed?

Hi @rotemtam, sure!

Sorry, I opened this draft PR before realising that it was almost 2:00 am 😅.

I'll update it shortly. I'm writing an example and a test just now!

PR #391, opened 2022-10-11
schemast: Support providing a base schema for a type

C-Deck:
* Add AddTypeWithBase method to provide a base schema when adding a type
    * Default AddType will use the ent.Schema
* Add BaseSchemaConfig for UpsertSchema and use AddTypeWithBase if the config is provided
* Unit testing for all of this

### DIFF
--- a/schemast/internal/mutatetest/policy/policytest.go
+++ b/schemast/internal/mutatetest/policy/policytest.go
@@ -1,0 +1,38 @@
+package policy
+
+import (
+	"context"
+
+	"entgo.io/ent"
+	"entgo.io/ent/privacy"
+)
+
+type (
+	FilterQueryRule func(context.Context, ent.Query) error
+
+	OnlyQuery interface {
+		OnlyID(ctx context.Context) (id string, err error)
+	}
+)
+
+func (f FilterQueryRule) EvalQuery(ctx context.Context, q ent.Query) error {
+	return f(ctx, q)
+}
+
+func (FilterQueryRule) EvalMutation(context.Context, ent.Mutation) error {
+	return nil
+}
+
+func IDFilterRule() ent.Policy {
+	return FilterQueryRule(func(ctx context.Context, q ent.Query) error {
+		IDQuery, ok := q.(OnlyQuery)
+		if ok {
+			_, err := IDQuery.OnlyID(ctx)
+			if err != nil {
+				return privacy.Deny
+			}
+		}
+
+		return privacy.Skip
+	})
+}
--- a/schemast/mutate.go
+++ b/schemast/mutate.go
@@ -24,9 +24,16 @@ import (
 )
 
 // Mutator changes a Context.
-type Mutator interface {
-	Mutate(ctx *Context) error
-}
+type (
+	Mutator interface {
+		Mutate(ctx *Context) error
+	}
+
+	BaseSchemaConfig struct {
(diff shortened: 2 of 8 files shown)


schemast/internal/mutatetest/policy/policytest.go
```diff
@@ -0,0 +1,38 @@
+package policy
```
a8m, line 1: I don't this code is needed for the test.

schemast/mutate.go
```diff
@@ -42,6 +49,7 @@ func Mutate(ctx *Context, mutations ...Mutator) error {
 // the type's Fields and Edges methods to return the desired fields and edges.
 type UpsertSchema struct {
 	Name        string
+	BaseConfig  *BaseSchemaConfig
```
a8m, line 52: "base schema" is not an official name we used in Ent (it's not written anywhere in the docs), but we still use it to describe the schema we embed as the "base" (usually, the zero one, `ent.Schema`), so I can understand why you used it. 

I think we can extend this option by letting users configure the [ast.TypeSpec](https://pkg.go.dev/go/ast#TypeSpec) definition instead using the mutators, and making it more generic. cc @rotemtam, what do you think?

a8m (commented):
Thanks for the contribution, @C-Deck.

Would you mind elaborating on your use case for `schemast`, how you use it, etc?

C-Deck:
> Thanks for the contribution, @C-Deck.
> 
> Would you mind elaborating on your use case for `schemast`, how you use it, etc?

We are dynamically generating the schema files from `json` definitions where the definition gets loaded into a struct that implements the following interface

```
	// SchemaGenerator describes the requirements to build a schema go file
	SchemaGenerator interface {
		// Name returns the name of the type
		Name() string

		// Annotations returns a list of schema annotations to be used by
		// codegen extensions.
		Annotations() []schema.Annotation

		// Edges returns the ent version of the edges
		Edges() []ent.Edge

		// Indexes returns the ent version of the indexes
		Indexes() []ent.Index

		// Fields returns the ent version of the fields
		Fields() []ent.Field
	}
```

Then, we are creating the upsert schema and doing the mutation like:

```
               upSchema := &schemast.UpsertSchema{
			Name:        schemaDef.Name(),
			Fields:      schemaDef.Fields(),
			Edges:       schemaDef.Edges(),
			Indexes:     schemaDef.Indexes(),
			Annotations: schemaDef.Annotations(),
		}

		err = upSchema.Mutate(loadCtx)
```

Unfortunately, we want all our models to have audit/specific mixins and a custom auth policy. When writing these by hand, we can embed the "base" schema with those. With `schemast`, you cannot have a custom emded schema and there is no capability to provide mixins or policy.

schemast/mutate.go
rotemtam: I think a nice way to add this is to add a `Mixin` field to `UpsertSchema` and use that to populate 

```go
func (User) Mixin() []ent.Mixin {
   // ...
}

```

somehow, WDYT?

C-Deck: @rotemtam That was how I saw it as well, but this way was a much simpler way to do the same thing for the time being. I'm willing to take a shot at it though.

C-Deck: Here is a brief attempt at mixin https://github.com/ent/contrib/pull/392

PR #394, opened 2022-10-23
fix(entproto): generate service implementation with custom id

vietdoan:
Signed-off-by: Viet Doan <vietdoanhp1996@gmail.com>

Related to https://github.com/ent/ent/issues/3027


entproto/internal/entprototest/ent/schema/service_with_id.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"entgo.io/contrib/entproto"
+	"entgo.io/ent"
+	"entgo.io/ent/schema"
+	"entgo.io/ent/schema/field"
+)
+
+// ServiceWithID holds the schema definition for the ServiceWithID entity.
+type ServiceWithID struct {
+	ent.Schema
```
rotemtam, line 26: Why not just expand `MessageWithID`? 

You can just add the `entproto.Service()` annotation on the existing entity.

entproto/internal/todo/ent/proto/entpb/entpb_grpc.pb.go
```diff
@@ -1,8 +1,4 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
-// versions:
```
rotemtam, line 2: notice this diff, make sure to use the same env as is used in CI (see readme for instructions), CI will regen all code and make sure no diff exists.

entproto/service.go
```diff
@@ -189,7 +189,7 @@ func (a *Adapter) genMethodProtos(genType *gen.Type, m Method) (methodResources,
 		outputName = "google.protobuf.Empty"
 		messages = append(messages, input)
 	case MethodList:
-		if !(genType.ID.IsInt() || genType.ID.IsInt64() || genType.ID.IsUUID() || genType.ID.IsString()) {
+		if !(genType.ID.Type.Type.Integer() || genType.ID.IsUUID() || genType.ID.IsString()) {
```
rotemtam, line 192: nice

entproto/internal/todo/ent/proto/entpb/entpb.proto
```diff
@@ -316,7 +316,7 @@ message Todo {
 }
 
 message User {
-  int64 id = 1;
+  uint32 id = 1;
```
rotemtam, line 319: @vietdoan doesnt this mean that this is a breaking change for all existing projects?

vietdoan: no, it's not. int32, uint32, int64, uint64, and bool are all compatible according to https://developers.google.com/protocol-buffers/docs/proto3

rotemtam: Yes I'm aware of that. This pertains to the proto wire format. The generated code will not be compatible though and will require changes from consumers. I'd like to avoid this if possible

vietdoan: @rotemtam I see! If they update dependencies, they inevitably have to manually update their code. However, isn't error during compile time not a big deal? Old consumers (using old protobuf schema) are still able to communicate with producers/servers (using new generated code) due to the compatibility mentioned above.

rotemtam (approved):
Thanks for the contribution!

PR #395, opened 2022-10-23
feat: add entgql to the schemast annotation

pepusz:
Annotations in the schemas package didn't support any entgql-related annotations. 
So I've added the annotations that I would like to use now. 

Hope I've used the right syntax.


schemast/annotation.go
```diff
@@ -186,6 +192,31 @@ func entSQL(annot schema.Annotation) (ast.Expr, bool, error) {
...
+			if input.IsCreate {
+				args = append(args, fnCall(selectorLit("entgql", "MutationCreate")))
+			} else {
+				args = append(args, fnCall(selectorLit("entgql", "MutationUpdate")))
+			}
+		}
+		c = fnCall(
+			selectorLit("entgql", "Mutations"), args...,
+		)
+	} else {
```
rotemtam, line 213: `entgql.Annotation` has many more fields. With the current implementation, users may provide input annotations with unsupported fields and we will silently ignore these options. I think when building infra code like this, correctness is of utmost importance. 

Can you either:
1. Add support for the missing fields
2. Return an error in case an unsupported field is used?

schemast/annotation.go
```diff
@@ -132,14 +134,18 @@ func protoEnum(annot schema.Annotation) (ast.Expr, bool, error) {
...
 	for k, v := range m.Options {
-		opts.Elts = append(opts.Elts, &ast.KeyValueExpr{
-			Key:   strLit(k),
-			Value: intLit(int(v)),
-		})
-	}
-	sort.Slice(opts.Elts, func(i, j int) bool {
-		return opts.Elts[i].(*ast.KeyValueExpr).Value.(*ast.BasicLit).Value < opts.Elts[j].(*ast.KeyValueExpr).Value.(*ast.BasicLit).Value
-	})
+		opts.Elts = append(
```
rotemtam, line 137: Thanks for all the formatting suggestions in this PR, but as they are not core to this change (and are a matter of style), can we discuss them in a separate PR?

rotemtam (commented):
Thanks for the contribution! Please find some comments on this PR.

schemast/annotation.go
pepusz: yea sure, just default intellij settings for me :) will revert these changes.

schemast/annotation.go
pepusz: Agree and will add all the missing fields in the upcoming days.

a8m:
@rotemtam, can you please re-review this? Thanks 🙏

schemast/annotation.go
```diff
@@ -15,6 +15,7 @@
 package schemast
 
 import (
+	"entgo.io/contrib/entgql"
```
a8m, line 18: Move this import down, please.

schemast/annotation.go
```diff
@@ -186,6 +192,31 @@ func entSQL(annot schema.Annotation) (ast.Expr, bool, error) {
...
 }
 
+func entGQL(annot schema.Annotation) (ast.Expr, bool, error) {
+	m := &entgql.Annotation{}
+	if err := mapstructure.Decode(annot, m); err != nil {
+		return nil, false, err
+	}
+	var c *ast.CallExpr
+	if m.MutationInputs != nil && len(m.MutationInputs) > 0 {
+		args := []ast.Expr{}
```
a8m, line 202:
```suggestion
		args := make([]ast.Expr, 0, len(m. MutationInputs))
```

pepusz:
I close now as i it is still heavily WIP will reopen later

PR #397, opened 2022-10-24
entgql: Limit fields selected from db when using CollectFields

samlof:


samlof:
todoTestSuite TestPaginationOrder breaks if remove priorityOrder field from the query... So have to fetch any fields that are used in ordering as well. Otherwise cursor will be wrong

entgql/internal/todo/todo_test.go
```diff
@@ -520,6 +524,179 @@ func (s *todoTestSuite) TestPaginationOrder() {
...
+			}))
+			s.Require().NotNil(rsp.Todos.PageInfo.StartCursor)
+			start := rsp.Todos.Edges[0]
+			s.Require().Equal(*rsp.Todos.PageInfo.StartCursor, start.Cursor)
+			s.Require().NotNil(rsp.Todos.PageInfo.EndCursor)
+			end := rsp.Todos.Edges[len(rsp.Todos.Edges)-1]
+			s.Require().Equal(*rsp.Todos.PageInfo.EndCursor, end.Cursor)
+			if i > 0 {
+				endCreatedAt, _ := time.Parse(time.RFC3339, end.Node.CreatedAt)
+				s.Require().True(startCreatedAt.Before(endCreatedAt) || startCreatedAt.Equal(endCreatedAt))
```
samlof, line 693: CreatedAt fields are empty for whatever reason so this fails

PR #401, opened 2022-10-27
entgql: support generate array for custom JSON type

nzlov:
Use:
```
		field.JSON("pause_times", []PauseTime{}).Annotations(
			entgql.Skip(entgql.SkipMutationCreateInput),
			entgql.Skip(entgql.SkipMutationUpdateInput),
		),
```

Gen:
```
  pauseTimes: [PauseTime!]!
```

Need `PauseTime` type to add declaration to your own graphql file.


entgql/schema.go
```diff
@@ -727,6 +727,9 @@ func (e *schemaGenerator) mapScalar(gqlType string, f *gen.Field, ant *Annotatio
 			// will generate an enum named "TodoStatus".
 			scalar = gqlType + scalar
 		}
+		if strings.HasPrefix(f.Type.String(), "[]") {
```
a8m, line 730: Use `RType.Kind` instead.

Also, can you please add test cases for this in `internal/todo`? Thanks

nzlov: Sorry.
RType is nil.

giautm:
why do you not use `entgql.Type('[PauseTime!]')`?

nzlov:
> why do you not use `entgql.Type("[PauseTime!]")`?

why must use?

entgql/schema.go
a8m: It's `nil` only for fields without GoType/JSON configured. Use: `f.Type.RType != nil && f.Type.RType.Kind = reflect.Slice`.

nzlov: ok

giautm: I'm not sure if there is a case when `Kind` is `reflect.Array`, but can we add it as a backup?

giautm:
Hi @nzlov, can you please fix the test? You need to run `go generate ./...` again in the EntGQL directory. Also, add the `custom` field as `Optional` to avoid breaking the currently test-cases.

entgql/schema.go
nzlov: ok

nzlov:
> Hi @nzlov, can you please fix the test? You need to run `go generate ./...` again in the EntGQL directory. Also, add the `custom` field as `Optional` to avoid breaking the currently test-cases.

ok

giautm:
I saw you forgot to update the gqlgen.yaml with the custom type. So, gqlgen generate another model for it.

nzlov:
> I saw you forgot to update the gqlgen.yaml with the custom type. So, gqlgen generate another model for it.

`gqlgen.yaml` `autobind` no work?

giautm:
> > I saw you forgot to update the gqlgen.yaml with the custom type. So, gqlgen generate another model for it.
> 
> `gqlgen.yaml` `autobind` no work?

https://github.com/ent/contrib/pull/401/files#diff-e367b846d59dfa604afc6c01ffbba09ca015cac77aed55b188658d2e6734a7d5
The config was missing for other example, like `gotype`, `pulid`, `uuid`.

Please try to fix all issue with PR, run `go test ./...` in local, and revert the change for go.mod, and entproto package

entgql/schema.go
```diff
@@ -727,6 +727,12 @@ func (e *schemaGenerator) mapScalar(gqlType string, f *gen.Field, ant *Annotatio
 			// will generate an enum named "TodoStatus".
 			scalar = gqlType + scalar
 		}
+		if f.Type.RType != nil {
+			switch f.Type.RType.Kind {
+			case reflect.Slice, reflect.Array:
+				scalar = "[" + scalar + "!]"
```
a8m, line 733: The "!" (not null) needs to be inferred from the type. Unlike `T`, `*T` is nullable.

a8m: @giautm, your thoughts on it? Users can always fallback to the `Type()` option.

giautm: > The "!" (not null) needs to be inferred from the type. Unlike `T`, `*T` is nullable.

agree.

nzlov:
I can't do the rest of my work because of the coronavirus

giautm:
> I can't do the rest of my work because of the coronavirus

Oh, Hope you getting better soon. I will take care the PR from here.

nzlov:
> > I can't do the rest of my work because of the coronavirus
> 
> Oh, Hope you getting better soon. I will take care the PR from here.

Thx.
I back.

entgql/schema.go
nzlov: How to fix?

nzlov:
@giautm Did the test fail because the tool version was too high and needed to be downgraded?

giautm:
Just revert the change in entproto

giautm:
Thank @nzlov for your contribution and patience. Wish you always healthy, and to make many contributions to the community in the future. Best regards, ❤️

giautm (approved):

nzlov:
@giautm thx.

kbumsik:
Wow I just needed this feature and was thinking to contribute it without knowing this PR.
Thank you very much for your work!

PR #206, opened 2022-01-16
Support multiple order field with entgql Paginate

yagipy:
closes https://github.com/ent/ent/issues/2198

PR #412, opened 2022-12-04
Support multiple order by for entgql

artificial-james:
Attempts to close https://github.com/ent/ent/issues/2198

Builds upon (and hopefully replaces) https://github.com/ent/contrib/pull/206

Should address concern (https://github.com/ent/contrib/pull/206#discussion_r788234502) from previous PR.

### DIFF
--- a/entgql/internal/todo/ent.graphql
+++ b/entgql/internal/todo/ent.graphql
@@ -83,7 +83,7 @@ type Category implements Node {
     last: Int
 
     """Ordering options for Todos returned from the connection."""
-    orderBy: TodoOrder
+    orderBy: [TodoOrder!]
 
     """Filtering options for Todos returned from the connection."""
     where: TodoWhereInput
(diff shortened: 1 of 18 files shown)


timoha:
Would this work for ordering by some 1-to-1 edge's field? For example, if I have schema like this:
```
type Patient implements Node {
  id: ID!
  dateOfBirth: Time!
  account: Account!
}
type Account implements Node {
  id: ID!
  firstName: String!
  lastName: String!
}
type Query {
    patients(
    after: Cursor
    first: Int
    before: Cursor
    last: Int
  ): PatientConnection
}
```
And I want to query patients by account's last name?

giautm:
hello @artificial-james and @timoha, this feature was implemented in https://github.com/ent/contrib/pull/442, please upgrade to the latest version to use. Thank you for your contribution. 🙇‍♂️

PR #414, opened 2022-12-08
entproto: support field.Ints

ducnguyen96:
Similar to  #285 

For converting pb `[]int64` to `[]int` I added `helpers` to `entproto` by using embed. Not so clever but that's the only solution I know.

Other part is just copy/paste from #285 

@rotemtam


entproto/adapter.go
```diff
@@ -522,6 +522,12 @@ func extractJSONDetails(f *gen.Field) (fieldType, error) {
 			repeated:  true,
 		}, nil
 	}
+	if f.Type.Ident == "[]int" {
```
a8m, line 525: Change this code to `switch`?

a8m (commented):
@rotemtam, please review this.

entproto/adapter.go
rotemtam: yes please @ducnguyen96

entproto/cmd/entproto/helpers.txt
```diff
@@ -0,0 +1,21 @@
+package entpb
+
+func toRepeatedInt(repeatedInt64 []int64) []int {
```
rotemtam, line 3: the way to add more code to the generated code is by extending the templates found here:
https://github.com/ent/contrib/blob/master/entproto/cmd/protoc-gen-entgrpc/template/to_ent.tmpl
https://github.com/ent/contrib/blob/master/entproto/cmd/protoc-gen-entgrpc/template/to_proto.tmpl

entproto/cmd/entproto/main.go
```diff
@@ -38,4 +43,21 @@ func main() {
 	if err := entproto.Generate(graph); err != nil {
 		log.Fatalf("entproto: failed generating protos: %s", err)
 	}
+
+	// open output file
+	fo, err := os.Create("proto/entpb/helpers.go")
```
rotemtam, line 48: use templates please ;-)

rotemtam (changes requested):
thanks for the contribution. see some comments and suggestions,
also notice CI is red, tips on solving the issues there can be found:
https://github.com/ent/contrib/tree/master/entproto#contributing

entproto/adapter.go
ducnguyen96: updated

entproto/cmd/entproto/main.go
ducnguyen96: used template

ducnguyen96:
> thanks for the contribution. see some comments and suggestions, also notice CI is red, tips on solving the issues there can be found: https://github.com/ent/contrib/tree/master/entproto#contributing
Thank you for the tips. I run `go test ./...` and seem there is no error anymore. Please approve CI to check.

Btw, this is my first contribution on open-source community ever. I'll be very happy if I can contribute a tiny piece of code to this meaningful project so any guide will be very much appreciated.

PR #415, opened 2022-12-09
bc: entproto: prefix pb enums with their field name

rotemtam:
Related to https://github.com/ent/ent/issues/3063

cc @ivanvanderbyl 

This is a ~breaking change to the generated code, but the wire format is unaffected. Clients using old-generated code can keep working with new servers and vice versa. 

To opt-out of this behavior (and use the previous one) users can use `entproto.OmitFieldPrefix()`

```
		field.Enum("omit_prefix").
			Values("foo", "bar").
			Annotations(
				entproto.Field(103),
				entproto.Enum(
					map[string]int32{
						"foo": 1,
						"bar": 2,
					},
					entproto.OmitFieldPrefix(),
				),
			),

```


giautm (approved):

a8m (approved):

ivanvanderbyl:
This is great, thanks @rotemtam 😍

PR #422, opened 2022-12-31
entgql: prefer edge schema over normal edge rel

tankbusta:
Edge Schema's are ***awesome*** and allow you to store metadata about a relationship (like when 2 user's first became friends) but the gql generator places both edges (friends and friendship) inside the node. However, the friendship edge looses the ability to filter. I noticed it creates the `WhereInput` fields inside the Go model but not GraphQL

https://github.com/ent/contrib/blob/7ed77ac9d07696176d7dfd80fd4cb3e8a1334582/entgql/internal/todo/ent/gql_where_input.go#L683-L689

This PR adds support for including the missing `WhereInput` fields in the schema and dropping the redundant edge on the nodes that have an edge schema. I would consider this a **breaking change** as it removes an edge that exists today but I struggled wondering how we could support backwards compatibility so I'm open to suggestions.

 I will probably need to run the Code Generator on all the examples after we've discussed this a bit more in-depth.


tankbusta:
@a8m @giautm Are we able to get some 👀 on this? Looks like the CI failure is a GitHub rate limit issue

entgql/annotation.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      http://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
```
a8m, line 7: @tankbusta, there are many changes like this that are irrelevant to this PR. Would you mind reverting them so it will be easier to review? Thanks

cc @giautm

giautm:
@tankbusta would you mind if I push to your branch? Look like it contains a lot of unrelated changes from the master.

tankbusta:
Yes please go ahead! I tried to rebase with master but I apparently messed up 😟

giautm:
@tankbusta: Can you help me fix broken test cases?

Ps: To avoid conflicts when you pull the code, please delete the local branch and fresh pull.

tankbusta:
I'll take a look. Interestingly, these tests didn't fail when I submitted

tankbusta:
Alright, should be fixed! Sorry about that - the issue was I didn't run the code generator when I originally submitted it because I wanted it to be easier for y'all to review. When you ran the code generator, it broke the test due to change this PR makes.

The test is a perfect example of what this PR changes; allowing fields that are added on an edge schema to be queryable (and filtered on) via GraphQL

tankbusta:
Anything I can do @giautm to help move this along?

PR #423, opened 2023-01-02
entoas: handle nillable fields

masseelch:
Thid PR adds support for `Optional()` and `Nillable()` fields. There are three cases to handle:

1. `Optional()`
This field has to be non-required in a POST request body and the model itself.

2. `Nillable()`
This field has to be required in a request body, but the result (model) might contain a (database) NULL value.

3. `Optional()` and `Nillable()`
Same as 1.


giautm (approved):
LGTM,

PR #428, opened 2023-01-20
entgql: fixed panic with latest ent

giautm:
Fixed https://github.com/ent/ent/issues/3256, https://github.com/ent/ent/issues/3265


a8m (approved):

brettinternet:
I'm trying to uncover a related issue with the pulid example and a "runtime error: invalid memory address or nil pointer dereference". This changeset is very difficult to explore with so many whitespace changes.

PR #430, opened 2023-01-26
entgql: fix problem with types that the underlying type is array/slice

crossworth:
This fixes: https://github.com/ent/ent/issues/3277

To solve the problem, we check the name of the type, if it's a slice/array, the name returned is empty.
https://go.dev/play/p/0CYtKzVOQxT

@giautm 

--- 
~~Something seems off on this PR, I don't know why there is a diff of `+1,402 −44,665`.~~


giautm (approved):
Thank so much!

PR #441, opened 2023-02-22
change collectField parameter name

tsingsun:
collectField parameter named op, easy to conflict with the receiver such as (OrganizationPolicy -> op )


a8m (approved):
Thanks, @tsingsun 🙏🏻

PR #444, opened 2023-02-23
fix multiOrder template

amrnt:
I think the multi order template was tested on category so `c` was hard coded.

I fix the template as well as added the annotation to `Group`.


a8m (approved):
Thanks, @amrnt 🚀

PR #446, opened 2023-03-02
add proto3 optional support

Adphi:


mikhail-eremin (approved):
Need this feature. Looks good

dizzyfool:
As I understand it, now instead of, for example,
```
google.protobuf.StringValue opt_field = 3;
```
it will be
```
optional opt_field = 3;
```

which will lead to loss of backward compatibility, as the generated code will change from
```go
optField *wrapperspb.StringValue
```
to
```go
optField *string
```
However, since wrappers are redundant in this particular case, I believe it's worth accepting this pull request.

Encorpluptit:
Any news on this ? Can this be merged ?
Would love to use optional field when generating schema with protoc-gen-ent !

jhoward-lm:
+1 for this

PR #452, opened 2023-03-08
entgql: Fix duplicate import present in `gql_where_input.go`

Adam-Mustafa:
Fixes https://github.com/ent/ent/issues/3373

* Removes the unnecessary call to template "import/types" since it was just creating an empty line.

### DIFF
--- a/entgql/template/where_input.tmpl
+++ b/entgql/template/where_input.tmpl
@@ -20,7 +20,6 @@ in the LICENSE file in the root directory of this source tree.
 import (
     "{{ $.Config.Package }}/predicate"
 	{{- range $n := $gqlNodes }}
-        {{ template "import/types" $n }}
 		"{{ $.Config.Package }}/{{ $n.Package }}"
 	{{- end }}
 )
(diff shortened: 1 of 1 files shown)


entgql/template/where_input.tmpl
```diff
@@ -20,7 +20,6 @@ in the LICENSE file in the root directory of this source tree.
 import (
     "{{ $.Config.Package }}/predicate"
 	{{- range $n := $gqlNodes }}
-        {{ template "import/types" $n }}
```
a8m, line 23: This template imports any packages required by custom `GoType`s. See https://github.com/ent/ent/blob/master/entc/gen/template/import.tmpl#L45-L57. 

Maybe change it like this?
```diff
-         {{ template "import/types" $n }}
+         {{- template "import/types" $n }}
```

a8m (approved):
Let's give it a try and see how it goes.

a8m:
Thanks for the contribution, @Adam-Mustafa.

PR #455, opened 2023-03-14
fix(entgql): check fields RType for Map and use Map scalar if possible

tankbusta:
If you create an entity with a JSON field...

```go
type Example struct {
	ent.Schema
}

func (Example) Fields() []ent.Field {
	return []ent.Field{
		field.JSON("metadata", map[string]interface{}{}).
			Optional(),
	}
}
```

You will get an error when running code generation

> 2023/03/13 19:21:39 running ent codegen: field(metadata): entgql: json type not implemented

You can resolve this today by adding the annotation `entgql.Type("Map")` but I didn't see any mention that this was required so I made the code a little bit smarter by checking if it's of `map[string]interface {}` and using the built-in `Map` scalar. If you do not set this, the error is made smarter by recommending that you set `entgql.Type`.


a8m:
Thanks for the contribution, @tankbusta 🙏🏻

Can you add a simple test for this? A field in one of these schemas: https://github.com/ent/contrib/tree/master/entgql/internal/todo/ent/schema

a8m (approved):
💯

entgql/schema.go
```diff
@@ -773,6 +773,13 @@ func (e *schemaGenerator) mapScalar(gqlType string, f *gen.Field, ant *Annotatio
 				case "[]string":
 					scalar = "[String!]"
 				}
+			case reflect.Map:
+				if f.Type.RType.Ident == "map[string]interface {}" {
```
giautm, line 777: the spacing between `interface` and `{}` really weird. But the test passed

giautm:
Thank @tankbusta for your contribution. Good to see your here again, <3

entgql/schema.go
tankbusta: Yeah, I was confused when I saw that at first but I think it's just how the `reflect` package identifies `interface{}`. You can see the same format in some of their tests: https://github.com/golang/go/blob/master/src/reflect/all_test.go#L713

PR #457, opened 2023-03-14
feat(entgql): Allow comments on edge fields

tankbusta:
If you add a comment to a field in the Ent schema, the comment will automatically carry over to the graphql schema. This behavior does not work on edges.

Of note, it may make sense to add a new `entgql.Comment("gql description here")` annotation that allows you to override the comment you have in your schema. The comment specified in the annotation would be used in the GraphQL schema; falling back to the comment added to the Field/Edge if the annotation was not provided. Thoughts?


giautm (approved):

giautm:
@tankbusta: thank you for your contribute. <3

PR #458, opened 2023-03-16
feat(entproto): make toProto{{ .EntType.Name }} functions public

mutsuki333:
In our project structure, we prefer to have our .proto and rpc implementations grouped in different modules, and importing the `entpb.proto` individually.

But the generated services now has the toProto functions private, and I wish to make it public.
Are there any concerns why these functions are designed to be private in the first place?


kbumsik:
I also need this PR. Because of the problem mentioned in this PR I am forced to put our protobuf integration code in the same folder.

Thank you for making this PR.

PR #459, opened 2023-03-16
entgql: only generate clear fields for optional edges

Kranex:
fixes the first part of ent/ent#3396

prevents the generation of the clear field in mutationInputs when an edge is not optional:

```go
func (Foo) Edges() []ent.Edge {
	return []ent.Edge{
		edge.To("required_bar", Bar.Type).
			Unique().
			Required(),
		edge.To("optional_bar", Bar.Type).
			Unique(),
	}
}
```

```go
type CreateFooInput struct {
	RequiredBar               some.ID
	OptionalBar                *some.ID
}

type UpdateFooInput struct {
	RequiredBar               *some.ID
	OptionalBar                *some.ID
	ClearOptionalBar       bool
}
```


giautm:
Hello @Kranex, thank you for your contribution. Please update GQL generate to remove the field too.

Kranex:
> Hello @Kranex, thank you for your contribution. Please update GQL generate to remove the field too.

@giautm done! I've also tried regenerating the examples in `/internal` but I don't think any of them are using generated mutation inputs with required edges.

[Additionally Ent seems be generating Clear methods for required edges in the `mutation.go`].(https://github.com/ent/ent/blob/055d367984badd8218e296cfac67475785651e75/entc/gen/template/builder/mutation.tmpl#L328) is this also something that would need changing?

Kranex:
@giautm is there anything that I still need to do for this?

giautm:
@Kranex: hello, can you add simple example to demostration the change, also use it as test?

a8m:
> @giautm is there anything that I still need to do for this?

Thanks for the contribution, @Kranex. As @giautm mentioned, pleass add a test case in `todo_test.go` and we can merge it. See: https://github.com/ent/contrib/blob/master/entgql/internal/todo/todo_test.go#L1875

Kranex:
@a8m only the Friendship schema has required edges but it is not using generated create and update inputs.

Would it be ok to enable the generated inputs for the friendship schema, even though you'd arguably never use them directly as the User schema creates them via the `through` builder?

entgql/internal/todo/todo_test.go
```diff
@@ -1907,6 +1907,39 @@ func TestMutation_ClearChildren(t *testing.T) {
...
+			ID string
+		}
+	}
+	err := gqlc.Post(`
+	mutation clearFriend($id: ID!){
+		updateFriendship(id: $id, input: {clearFriend: true}) {
+			id
+		}
+	}
+	`, &rsp, client.Var("id", friendship.ID))
```
Kranex, line 1938:
```suggestion
	var rsp struct {
		UpdateFriendship struct {
			ID string
		}
	}
	err := gqlc.Post(`
	mutation clearFriend($id: ID!){
		updateFriendship(id: $id, input: {clearFriend: true}) {
			id
		}
	}
	`, &rsp, client.Var("id", ""))
```

Kranex:
@a8m @giautm I've gone ahead and added an `updateFriendship` mutation, along with a test to confirm that the `clear` field is not generated for the required `friend` edge.

The test currently creates two new users and a friendship between them, but given that the request is expected to throw a schema validation error [this suggested change would also be valid](https://github.com/ent/contrib/pull/459/files#r1160871412).

a8m (approved):
💯

a8m:
Thanks for the contribution, @Kranex.

PR #461, opened 2023-03-17
feat(protoc-gen-ent): google.protobuf.Timestamp to field time

exuan:


rotemtam (changes requested):
Hey @exuan 
Thanks for your contribution.
Please add an assertion on `TestFieldModifier` that the field is generated correctly.
Thanks

exuan:
> Hey @exuan Thanks for your contribution. Please add an assertion on `TestFieldModifier` that the field is generated correctly. Thanks

Okay, I recommitted [`created_at`  assertion](https://github.com/ent/contrib/blob/00e38a300364c576012024b7dd01551a4fae4171/entproto/cmd/protoc-gen-ent/main_test.go#L59)

PR #449, opened 2023-03-04
protoc gen support complex directory

tsingsun:
If your project organization directory is more complex, such as the Ent schema directory and the target directory and proto directory are separate. ent proto is not support it. the new PR resolver it

like such as:
```
- api
- - proto    // ent proto generated file
- ent          // ent generated
- graph
- - schma    // the ent schema
```

PR #462, opened 2023-03-18
entproto: configure target dir with extension api

rotemtam:
Use `entproto` using the Extension API. 

Adds control for target dir similar to #449.


entproto/extension.go
```diff
@@ -22,12 +22,70 @@ import (
...
 
+	"entgo.io/ent/entc"
 	"entgo.io/ent/entc/gen"
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/desc/protoprint"
 	"go.uber.org/multierr"
 )
 
+// ExtensionOpt is an option for the entproto extension.
+type ExtensionOpt func(*Extension)
```
a8m, line 33: Let's keep the naming the same as entgql? https://github.com/ent/contrib/blob/master/entgql/extension.go#L41
```suggestion
type ExtensionOption func(*Extension)
```

entproto/extension.go
```diff
@@ -22,12 +22,70 @@ import (
...
+
+// New returns a new Extension configured by opts.
+func New(opts ...ExtensionOpt) *Extension {
```
a8m, line 36: Ditto: https://github.com/ent/contrib/blob/master/entgql/extension.go#L205
```suggestion
func NewExtension(opts ...ExtensionOpt) *Extension {
```

entproto/extension.go
```diff
@@ -36,12 +94,11 @@ import (
 //	    entproto.Hook(),
 //	  },
 //	})
+//
+// Deprecated: use Extension instead.
```
a8m, line 98: Use `Extension` then to ensure we don't maintain this code twice:

```
NewExtension().hook()
```

entproto/extension.go
a8m: You may also want to change the function to return an error `(*Extension, error)`. You don't need this now, but it will be a pain to change it in the future in case we need it.

a8m (approved):

PR #463, opened 2023-03-19
entgql: add support for query fields selection

a8m:
Note that if the schema was extended, an unrecognized field was queried, entgql skips the fields selection because we cannot determine if a non-selected field is used in custom resolvers.

Wait with the review until I add tests to this PR.

### DIFF
--- a/entgql/template/collection.tmpl
+++ b/entgql/template/collection.tmpl
@@ -38,10 +38,22 @@ func ({{ $receiver }} *{{ $query }}) CollectFields(ctx context.Context, satisfie
 	return {{ $receiver }}, nil
 }
 
-func ({{ $receiver }} *{{ $query }}) collectField(ctx context.Context, opCtx *graphql.OperationContext, field graphql.CollectedField, path []string, satisfies ...string) error {
+func ({{ $receiver }} *{{ $query }}) collectField(ctx context.Context, opCtx *graphql.OperationContext, collected graphql.CollectedField, path []string, satisfies ...string) error {
 	path = append([]string(nil), path...)
-	{{- if $collects := fieldCollections (filterEdges $node.Edges (skipMode "type")) }}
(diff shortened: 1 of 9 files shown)


entgql/template/collection.tmpl
```diff
@@ -38,10 +38,13 @@ func ({{ $receiver }} *{{ $query }}) CollectFields(ctx context.Context, satisfie
...
 
-func ({{ $receiver }} *{{ $query }}) collectField(ctx context.Context, opCtx *graphql.OperationContext, field graphql.CollectedField, path []string, satisfies ...string) error {
+func ({{ $receiver }} *{{ $query }}) collectField(ctx context.Context, opCtx *graphql.OperationContext, collected graphql.CollectedField, path []string, satisfies ...string) error {
 	path = append([]string(nil), path...)
-	{{- if $collects := fieldCollections (filterEdges $node.Edges (skipMode "type")) }}
-		for _, field := range graphql.CollectFields(opCtx, field.Selections, satisfies) {
+	{{- $fields := filterFields $node.Fields (skipMode "type") }}
+	{{- $collects := fieldCollections (filterEdges $node.Edges (skipMode "type")) }}
+	{{- if or $collects $fields }}
+		selectedFields, fieldSeen, unknownSeen := make([]string, 0, len({{ $node.Package }}.Columns)), make(map[string]struct{}, len({{ $node.Package }}.Columns)), false
```
giautm, line 46:
```suggestion
	var (
		selectedFields = make([]string, 0, len({{ $node.Package }}.Columns))
		fieldSeen = make(map[string]struct{}, len({{ $node.Package }}.Columns))
		unknownSeen = false
	)
```

giautm: I love to use var here. :D

giautm (approved):
💯

PR #469, opened 2023-03-24
added directives to create and update mutations

lucvankessel:
This is the fix for [this issue](https://github.com/ent/ent/issues/3407) i created a few days ago.
Right now however it will add it to the type, create and update mutation. 
And i would like to have the option in the future to skip a directive on a type or mutation (create/update) is that something you also agree with adding?


lucvankessel:
Hey @a8m,
Do you have additional questions about this mr? Or are there tests that need to be written still? would love to know if there is anything i can do to help get this merged.

giautm:
Hello @lucvankessel, thank you for your contribution and patience here.

I and Ariel already have a discussion about this PR. We think we need some changes to this PR before accepting this. The Directives annotation should work on scoped objects (Object `Type` or Object `Input`). So it avoids breaking change with current users.

Bellow is my suggestion API.
```go
entgql.Directives(entgql.Deprecated("Use `description` instead.")).
  OnTypes("Todo"). // Default
  OnInputs("CreateTodoInput", "CreateTodoUpdate")
```

cc @a8m

lucvankessel:
@giautm thanks for the response.
Is there a reason why all the directives have the same inputs and types instead of deciding that on a per directive basis?
Cc @a8m

lucvankessel:
Hey @giautm and @a8m,
Sorry for the bump again.
After looking into it for a bit i woud like your feedback on the following API example
```
entgql.Directives(entgql.Deprecated("User `description` instead.")).OnCreateInput().OnUpdateInput().SkipType()
```
This (i think) will do the same thing you suggested but without the typed types and inputs.
It also makes sure it is backwards compatible and still enable the directives on the inputs i want.
It is still not the most perfect solution (my preference still goes out to using the existing skip mode type and making a skipDirectives field for it in the annotation, or even better having it on the directive itself so some directives can be on the update, some on the create and some on the type. for these skip fields a default value would be set to make sure it behaves like it does now)

lucvankessel:
Could i get a reply from one of the devs on my comments? @giautm @a8m

syssam:
@lucvankessel, @giautm how about https://github.com/ent/ent/issues/3588

PR #481, opened 2023-04-13
entgql: support ordering by edge count/field

a8m:
Next PRs will include adding multiple terms to edges, but this requires some internal logic for merging these terms into one join per relation.


giautm (approved):
:100: hurray!!!

PR #482, opened 2023-04-14
entproto: add comment in generate.go to mark the file as generated

jamslinger:
I ran into the issue that my linter was complaining about the entproto-generated `generate.go` file due to a missing new line. This RP adds the new line but I think the file should also contain the `// Code generated by entproto, DO NOT EDIT.` comment which is also present in all other generated files.


rotemtam:
Hey

entproto generates this file initially but once it exists it does not touch it. This is because we expect users will tweak the protoc invocation to their needs. Ergo, we decided to omit the codegen comment, much like if you run 

go run entgo.io/ent/cmd/ent new User

ent/schema/user.go will not contain a codegen header comment.

Closing this, but do reopen if you have any further questions or reservations

jamslinger:
Thanks for your comment, regarding the `generated` comment that makes sense. Would you agree to add the second `\n` after the `package` line so that the file is properly formatted from the beginning on? We use the generated file in our CI directly and I'd like to avoid any special linter config to exclude the file.

PR #487, opened 2023-04-20
entproto: support enum values with special characters

jamslinger:
This fixes: https://github.com/ent/ent/issues/3498


rotemtam (changes requested):
I like this . Please handle the case of two enums that normalize to a identical string with an informative error?

jamslinger:
> I like this . Please handle the case of two enums that normalize to a identical string with an informative error?

Thanks for your feedback, that's a good point. I assume you mean two values within a single enum right? I updated the PR accordingly and also added a test.

I also factored out a [`entproto.NormalizeEnumIdentifier`](https://github.com/jamslinger/contrib/blob/enum-value-normalization/entproto/enum.go#L142) function so that we don't rely on the same normalization logic being implemented in multiple different places. It is still being defined twice though, in `entproto.NormalizeEnumIdentifier` as well as in [the enum template](https://github.com/jamslinger/contrib/blob/enum-value-normalization/entproto/cmd/protoc-gen-entgrpc/template/enums.tmpl#L11).

I also have a question: We are not being consistent in error messages with strings vs. quoted strings and I assume we are stuck with this for backwards compatibility?

rotemtam (approved):
Thanks for the contribution @jamslinger !

rotemtam:
@jamslinger CI is red, can you check ?

jamslinger:
I did a `go mod tidy` and it should be fixed now.

rotemtam:
Thanks for the contribution @jamslinger !

jamslinger:
Thanks for the support @rotemtam :)

PR #497, opened 2023-05-03
Add failing test when have a deeper fragment(s)

amrnt:


entgql/internal/todo/todo_test.go
```diff
@@ -2732,3 +2732,67 @@ func TestSatisfiesFragments(t *testing.T) {
...
+		fragment Child2 on Category {
+			id
+			text
+		}
+
+		fragment Child1 on Todo {
+			text
+			category {
+				id
+				... Child2
```
amrnt, line 2771: having `text` under `Child1.category` will make it pass as well

a8m:
Closed with https://github.com/ent/contrib/pull/498. Thanks a lot for reporting this

PR #498, opened 2023-05-03
entgql: collect interface/union selection fields of unique edges

a8m:


amrnt:
@a8m @giautm  Awesome stuff 💯 🎉

PR #501, opened 2023-05-09
ci: avoid rate limit

Zxilly:
prevent https://github.com/ent/contrib/actions/runs/4910216651/jobs/8779388684


rotemtam (commented):
Thanks for the suggestion, I prefer not to provide the repo secret to a 3rd party repo that I do not control. Closing

PR #503, opened 2023-05-09
fix: ensure graph target has a path

rdeusser:
`g.Config.Target` always seems to be an empty string.

Fixes https://github.com/ent/ent/issues/3467

### DIFF
--- a/entproto/extension.go
+++ b/entproto/extension.go
@@ -110,6 +110,15 @@ func Generate(g *gen.Graph) error {
 }
 
 func (e *Extension) generate(g *gen.Graph) error {
+	if g.Config.Target == "" {
+		abs, err := filepath.Abs("./")
+		if err != nil {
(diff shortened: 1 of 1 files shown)


rdeusser:
@rotemtam No rush, but do you have an idea of when you might be able to look at this?

Capable-Disk4147 (approved):
It seems good to me. I wish I could have configured the config by a CLI flag or something as a temporary workaround.

Capable-Disk4147:
@rotemtam grpc support is completely broken for ent until this is merged and has been for over a month. If you could at least comment to say you're aware of this, that would be appreciated

earljwagner:
I'm also unable to use grpc until this is merged. If @rotemtam is unavailable, would it be possible to have someone else review it?

rotemtam:
hey all
merged #521 instead with a slightly different approach and tests. 
Apologies to all for the very late response. 
Closing

PR #510, opened 2023-05-24
entproto: support json for integer type slices

jamslinger:
This implements https://github.com/ent/ent/issues/3559


rotemtam:
Thanks for the contribution @jamslinger!

Will review this soon.

PR #512, opened 2023-05-28
entgql: fixed Paginate query for count

giautm:
Related https://github.com/ent/ent/issues/2984

### DIFF
--- a/entgql/template/pagination.tmpl
+++ b/entgql/template/pagination.tmpl
@@ -632,7 +632,7 @@ func ({{ $r }} *{{ $name }}) ToEdge(order *{{ $order }}) *{{ $edge }} {
 		hasPagination := after != nil || first != nil || before != nil || last != nil
 		if hasPagination || ignoredEdges {
 			c := {{ $r }}.Clone()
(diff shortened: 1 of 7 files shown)


entgql/template/pagination.tmpl
```diff
@@ -631,7 +631,10 @@ func ({{ $r }} *{{ $name }}) ToEdge(order *{{ $order }}) *{{ $edge }} {
 	if hasCollectedField(ctx, totalCountField) || hasCollectedField(ctx, pageInfoField) {
 		hasPagination := after != nil || first != nil || before != nil || last != nil
 		if hasPagination || ignoredEdges {
-			if conn.TotalCount, err = {{ $r }}.Clone().Count(ctx); err != nil {
+			c := {{ $r }}.Clone()
+			// We don't need to select the fields of the node, since we only need the count.
```
a8m, line 635: Change it to a template comment instead? There’s no need having this in all generated code.

a8m, line 635:
```suggestion
			{{- /* Clear the selection fields before counting to avoid generating invalid queries. */}}
```

a8m (approved):

PR #529, opened 2023-07-10
Include `fieldSeen` if fieldCollections returns non-empty 

Adam-Mustafa:


giautm:
Hello, can you describe what problem you're trying to resolve here? so I can support to resolve it faster. Thank you.

PR #531, opened 2023-07-11
 fix tmpl package alias

tsingsun:
use some  package import code as ent latest commit.

[ent tmpl](https://github.com/ent/ent/blob/c46cf6317b25707f86bd568acbaefd083c2c175f/entc/gen/template/client.tmpl#L61)

```
line61   {{ $n.PackageAlias }} "{{ $n.Config.Package }}/{{ $n.PackageDir }}"
```
and Graph.Type
```
// Package returns the package name of this node.
func (t Type) Package() string {
	if name := t.PackageAlias(); name != "" {
		return name
	}
	return t.PackageDir()
}
```

the package name use alias first.

This PR avoid Schema Node use alias name but gql code use Schema name

### DIFF
--- a/entgql/template/where_input.tmpl
+++ b/entgql/template/where_input.tmpl
@@ -20,7 +20,7 @@ import (
     "{{ $.Config.Package }}/predicate"
 	{{- range $n := $gqlNodes }}
         {{- template "import/types" $n }}
-		"{{ $.Config.Package }}/{{ $n.Package }}"
+        {{ $n.PackageAlias }} "{{ $n.Config.Package }}/{{ $n.PackageDir }}"
 	{{- end }}
 )
 {{ template "import" $ }}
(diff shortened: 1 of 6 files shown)


entgql/template/where_input.tmpl
```diff
@@ -20,7 +20,7 @@ import (
     "{{ $.Config.Package }}/predicate"
 	{{- range $n := $gqlNodes }}
         {{- template "import/types" $n }}
-		"{{ $.Config.Package }}/{{ $n.Package }}"
+        {{ $n.PackageAlias }} "{{ $n.Config.Package }}/{{ $n.PackageDir }}"
```
a8m, line 23: Fix the indentation, please.

a8m (commented):
Looks good, but see the small comment.

tsingsun:
In addition, I found that if exists aliases, it would cause `go import` to be abnormal. I adopted a rename Schema to prevent aliases from appearing.

Due to my tight work schedule, I haven't submitted this question yet.

PR #534, opened 2023-07-16
Override Type name for mutation input types

ysaakpr:


a8m (commented):
Thanks for the contribution, @ysaakpr. Please add an example usage+tests for it (See the `internal/todo` package).

PR #537, opened 2023-07-19
Privatize columns used in cursors predicates

amrnt:


a8m (commented):
PR looks good, I don't see any problems merging this. But are you able to reproduce this somehow with a test or just elaborate where does it fail?

amrnt:
Using multi order, It does fail when for instance having `Paginate`  chained to `Modify` with joins for example, and the order fields does exist in both (or more) tables.

The error was eg:
```
panic: [{"message":"ambiguous column name: status","path":["todosWithJoins"]}]
```

I added a test case

amrnt:
@a8m does the latest push satisfies the PR?

a8m (approved):
💯 

Thanks, @amrnt.

PR #538, opened 2023-07-25
fix: rename variable when iterating through p.order in toCursor template

Tantatorn-dev:
According to https://github.com/ent/ent/issues/3601, I've found this error too when my table's name starts with 'O'. 
It seems variables' name are conflicting; the argument of toCursor function and the one for iterating through p.order.
I've renamed the variable for iterating through p.order to 'po' in a template.


a8m:
Thanks for the contribution, @Tantatorn-dev. You need to run `go generate` on entgql.

Tantatorn-dev:
@a8m I've run go generate

a8m (approved):
Thanks for your contribution, @Tantatorn-dev.

PR #542, opened 2023-08-05
protoc-gen-entgrpc: Fix missing import when using GoType with basic field

rlebel12:
Hi, all! I've been having a great time using this library, though I recently hit a snag with gRPC service generation.

Currently, when specifying a `GoType` for a field of basic type (e.g. `field.Uint64`), the protobufs are generated correctly, and the gRPC service code is mostly correct, but the latter is missing an import statement for the package containing the custom type passed to `GoType`. This example schema demonstrates the issue:
```go
package schema

import (
    "entgo.io/contrib/entproto"
    "entgo.io/ent"
    "entgo.io/ent/schema"
    "entgo.io/ent/schema/field"
)

type CustomType uint64

type Example struct {
    ent.Schema
}

func (Example) Fields() []ent.Field {
    return []ent.Field{
        field.Uint64("custom_field").
            Annotations(entproto.Field(2)).
            GoType(CustomType(0)),
    }
}

func (Example) Annotations() []schema.Annotation {
    return []schema.Annotation{
        entproto.Message(),
        entproto.Service(),
    }
}

```
The generated service code `entpb_example_service.go` is missing the import of the `schema` package containing `CustomType`, but the actual type conversion taking place in the service is otherwise correct.

The proposed change copies an existing pattern (aside: maybe worth refactoring at this point) for detecting and setting constructors for fields with a `GoType` such that the final gRPC code includes the dependencies. If additional tests are needed, I'd be happy to give it a shot if someone could point out the right place for them.

Also, I thought I'd make an attempt at implementing a solution directly, but if this seems like a valid issue that perhaps would need more involved work, I could also file an issue to track it.


rlebel12:
One thing to note is that this change would cause service generation to fail when a model like the following is used, where `GoType` is called with an equivalent `ent.Field`:
```go
func (Example) Fields() []ent.Field {
	return []ent.Field{
		field.Uint64("test").
			Annotations(entproto.Field(2)).
			GoType(uint64(0)).
	}
}
```
The error:
```
--entgrpc_out: template execution failed: template: to_proto.tmpl:60:17: executing "field_to_proto" at <newConverter .Field>: error calling newConverter: runtime error: index out of range [1] with length 1
```
I don't know whether that's something that would ever be done/need to be supported in practice, but I wanted to call it out. If that functionality should be preserved, the check I implemented could be changed to be something like the following to compare against the underlying type (or perhaps something more idiomatic 😃):
```go
case efld.Type.Numeric() && efld.HasGoType() && efld.Type.RType.Name != efld.Type.RType.Kind.String():
```

rlebel12:
Hi @rotemtam, any chance of an update on this?

rotemtam (commented):
Hey,
It would be good to support basic GoTypes for protoc-ent-grpc. 
A few points: 
1. Yes, please add tests for this. A good way to do that is to add some fields of various types to the schema [here](https://github.com/ent/contrib/tree/master/entproto/internal/todo/ent/schema) and to add some test cases that verify they are used properly [here](https://github.com/ent/contrib/blob/master/entproto/internal/todo/ent/proto/entpb/user_service_test.go). Also test cases where the types are defined in the schema package and outside it.
2. If we support strings, lets support types that are convertible to basic go types.

rlebel12:
This has languished for a while without me giving it more attention, so I'm closing this for now. If desired, the next step to moving this forward would be to split out all the types in the `gotypes` test package into separate packages such that we thoroughly test that the various `Type` groups are imported as expected.

PR #543, opened 2023-08-11
added proposed directive changes

lucvankessel:
These are the changes i proposed in MR https://github.com/ent/contrib/pull/469   
I have added tests to show the backwards compatability. If they are unnecessary or need changing let me know.
I think it could be made prettier/sleeker but to ensure it is reliably backwards compatible this seemed like the best implementation.


syssam:
@lucvankessel any good news?

lucvankessel:
> @lucvankessel any good news?

@syssam no

@a8m would you be able to take a look?

PR #546, opened 2023-09-25
test: check custom predicate using sql join with generated predicate

michaelcaulley:
If using a SQL JOIN in a custom predicate in conjunction with an automatically generated predicate, the JOIN gets dropped from the generated SQL, and only the WHERE gets added.

See failing test as an example.

Error:
```
       	            	[{"message":"no such column: t1.status","path":["todos"]}]
```


michaelcaulley:
@a8m curious if this does look like an issue to you. I can try to find some time in the next week or two to dig more if so.

a8m (commented):
Hey! I think this was fixed in latest versions of ent.

Can you try updating ent and report if you still encounter issues with this?

michaelcaulley:
@a8m, I just pulled in the `master` branch of ent into this branch and the test passes. Looks like this issue is fixed. Thank you for following up. 

Can you check out the other issues I ran into?
https://github.com/ent/contrib/pulls/michaelcaulley

PR #547, opened 2023-10-03
feat: allow aggregate ordering of through table edges

michaelcaulley:
#Summary
* Add support for using `entgql.OrderField()` for edges that use a Through table.
* The OrderField string must be prefixed with the string value of the edge. For example:
```
	edge.To("friends", User.Type).
			Through("friendships", Friendship.Type).
			Annotations(entgql.RelayConnection()),
			Annotations(entgql.RelayConnection()),
```
would require the OrderField string to start with `FRIENDS_`

For the count aggregation the full string would be `FRIENDS_COUNT`.

* The changes skip the through table OrderBy and only use the original edge OrderBy.


entgql/template.go
```diff
@@ -443,12 +443,15 @@ func orderFields(n *gen.Type) ([]*OrderTerm, error) {
...
+	previousEdgeHasThroughTable := false
 	for _, e := range n.Edges {
 		name := strings.ToUpper(e.Name)
 		switch ant, err := annotation(e.Annotations); {
 		case err != nil:
 			return nil, err
-		case ant.Skip.Is(SkipOrderField), ant.OrderField == "":
+		case ant.Skip.Is(SkipOrderField): // skip edges marked with the skip annotation.
+		case ant.OrderField == "": // skip edges without the order field annotation.
+		case previousEdgeHasThroughTable: // orderBy is done on the edge, not the through table, so skip
```
michaelcaulley, line 454: Is this problematic in any cases? All the tests seem to pass for me.

michaelcaulley: It seems that for an edge with a through table that the list of edges has both the edge name and the through edge name. They appear in order, so we can skip the second through edge orderBy.

michaelcaulley:
Thanks for taking a look at this @giautm. I generated the code for the projects that I missed.

michaelcaulley:
@giautm anything I can help with to get this merged?

entgql/template.go
```diff
@@ -443,12 +443,19 @@ func orderFields(n *gen.Type) ([]*OrderTerm, error) {
...
 	for _, e := range n.Edges {
 		name := strings.ToUpper(e.Name)
 		switch ant, err := annotation(e.Annotations); {
 		case err != nil:
 			return nil, err
-		case ant.Skip.Is(SkipOrderField), ant.OrderField == "":
+		case ant.Skip.Is(SkipOrderField):
+			// skip edges marked with the skip annotation
+		case ant.OrderField == "":
+			// skip edges without the order field annotation
```
a8m, line 455: Revert this.

entgql/template.go
```diff
@@ -443,12 +443,19 @@ func orderFields(n *gen.Type) ([]*OrderTerm, error) {
...
+		case strings.HasSuffix(ant.OrderField, "_COUNT") &&
+			edgeNamesWithThroughTablesSet[ant.OrderField[:len(ant.OrderField)-6]] != nil:
```
a8m, line 457: Not sure I understand the logic here, but I think the current bug can be fixed on the template side.

michaelcaulley: It was looking like an order field was being created for both the edge and the join table, but it's only needed for one.

I think if it's done in the templates we'd have to update it in more than one spot, but perhaps that'd be the more correct way to go about it.

PR #548, opened 2023-10-18
entproto: when parsing fails, show detailed error

lazychanger:
fix https://github.com/ent/ent/issues/3795

### DIFF
--- a/entproto/adapter.go
+++ b/entproto/adapter.go
@@ -158,6 +158,12 @@ func (a *Adapter) parse() error {
 		}
 	}
 
+	if len(a.errors) > 0 {
+		for _, err := range a.errors {
+			return err
+		}
+	}
+
 	// Append the well known types to the context.
 	for _, wktPath := range wktsPaths {
 		typeDesc, err := desc.LoadFileDescriptor(wktPath)


lazychanger:
**old**
 `entproto: failed generating protos: entproto: failed parsing ent graph: proto: message field "entpb.Corporation.servers" cannot resolve type: "*.Server" not found`

**now**
`entproto: failed generating protos: entproto: failed parsing ent graph: entproto: default value for Enum pbfield "original_type" is "ip_hash", but the proto annotation pbfield 0 is "unspecified"`

rotemtam (commented):
It seems like there are two changes here? Are the changes to idType related to the error message?

lazychanger:
I'm sorry, it was an accident🤣. I will remove it

entproto/adapter.go
```diff
@@ -158,6 +158,12 @@ func (a *Adapter) parse() error {
 		}
 	}
 
+	if len(a.errors) > 0 {
+		for _, err := range a.errors {
+			return err
+		}
```
giautm, line 164:
```suggestion
        return a.errors[0]
```

PR #549, opened 2023-10-19
feat: add support for similarly named edges on different ent objects …

michaelcaulley:
## Summary
-  add support for similarly named edges on different ent objects to use a common edge orderBy field.
   - example: A Document schema with a edge to a User owner with order by field `owner_full_name` and a Project schema with an edge to a User owner with the order field `owner_full_name`, would cause a conflict in the generated OrderField of gql_pagination.go.
  
## The Fix
- For the OrderField struct, the type should always be the type of ent being sorted. In the above example, the generated OrderField struct will now be  `DocumentOrderFieldOwnerFullName` and `ProjectOrderFieldOwnerFullName`. And when dealing with the generated code to select the field of the edge, we should use the edge type.

### Changes (Not including generated code and test updates)
- [entgql/template.go](https://github.com/ent/contrib/pull/549/files#diff-6e7f2e0d40282a06096ab27365743cb1d2d7d2389d3ed941e5cac14a88e8373f)
- [entgql/template/pagination.tmpl](https://github.com/ent/contrib/pull/549/files#diff-bd3c2af6cb7cfbff02e6c1e42a207c8ccd2b38ded64ca5a988507cd69b4b2250)

## Test Plan
- added CREATED_BY_NAME order by edge to Group and Todo ents to verify that the code generation happens successfully


michaelcaulley:
@a8m Here's the pull request. I think we need to consider the edge owner name when creating the variable holding the order term.

michaelcaulley:
@giautm Can you help review this?

entgql/template.go
```diff
@@ -476,7 +476,7 @@ func orderFields(n *gen.Type) ([]*OrderTerm, error) {
 			terms = append(terms, &OrderTerm{
 				GQL:   ant.OrderField,
 				Edge:  e,
-				Type:  e.Type,
+				Type:  n,
```
a8m, line 479: `Type` is the ent type that owns the field. The previous version is correct. The problem is in the template that renders the code

a8m: The owner of this order term passed/attached here, and I assume this will solve the issue https://github.com/ent/contrib/blob/master/entgql/template.go#L397

michaelcaulley: I think to keep this Type value the same we'll have to add another field to this OrderTerm struct so that the template knows the type being sorted. Do you agree? I thought about that as a solution but I ended up thinking that the Type of the OrderType is the type of ent object being sorted, not the type of the edge. But perhaps that's not the way to think about it.

In the example of sorting a Todo by the Owner's name the old code would generate something like: 
```
GQL: OWNER_NAME,
Edge: User Edge
Type: User
Field: name
```
after the current change it'd be:
```
GQL: OWNER_NAME,
Edge: User Edge
Type: Todo
Field: name
```

In the case of sorting on a field, no edge, it'd be 
```
GQL: Text,
Edge: nil
Type: Todo
Field: text
```

If we keep this as is and instead add a new field to the OrderTerm struct, do you have any preference for what to call it? I don't think OwnerType is great because it's not necessarily the owner of the edge relationship. If we use the edge owner there is still a chance for conflict, we have to use the type of the object being sorted.

SortedType? EntType? Something else?

a8m:
Hey @michaelcaulley! I fixed that with https://github.com/ent/contrib/pull/565. I really sorry, I used `amend` accidentally instead of taking your commit (merge commit). I appreciate your contribution to ent and its community, and I hope you will have more opportunities to contribute soon.

Thanks again 🙏🏻

michaelcaulley:
No worries. Thanks for helping get the fix into the main branch.

PR #550, opened 2023-10-20
add json tag in entgql mutation_input template

ThinkontrolSY:


masseelch:
Hey! What is the use case for this? Omitempty?

a8m:
> Hey! What is the use case for this? Omitempty?

I am joining this question. gqlgen auto-binding works without these tags and those types are used mainly by the client side where MarshalJSON is not used - the server calls UnmarshalJSON only.

PR #551, opened 2023-10-28
Add json tags to gql_mutation_input

philip-peterson:
Should solve https://github.com/ent/ent/issues/3086

Similar to https://github.com/ent/contrib/pull/550

### DIFF
--- a/entgql/template.go
+++ b/entgql/template.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"text/template"
 	"text/template/parse"
+	"unicode"
 
 	"entgo.io/ent/entc/gen"
 	"entgo.io/ent/schema/field"
@@ -77,6 +78,7 @@ var (
 
 	// TemplateFuncs contains the extra template functions used by entgql.
 	TemplateFuncs = template.FuncMap{
+		"decap":               decap,
 		"fieldCollections":    fieldCollections,
 		"fieldMapping":        fieldMapping,
 		"filterEdges":         filterEdges,
@@ -170,6 +172,27 @@ func gqlUnmarshaler(f *gen.Field) bool {
 type fieldCollection struct {
 	Edge    *gen.Edge
 	Mapping []string
+}
+
+// decap takes a string and decapitalizes the first letter
+//
+//	FullName  => fullName
+//	ID  => iD
+//	user_id    => userID
+//	full-admin => fullAdmin
+func decap(s string) string {
+	var (
+		b strings.Builder
+	)
+	for i := 0; i < len(s); i++ {
+		r := rune(s[i])
+		if i == 0 {
+			b.WriteRune(unicode.ToLower(r))
+		} else {
+			b.WriteRune(r)
+		}
+	}
+	return b.String()
 }
 
 func fieldCollections(edges []*gen.Edge) ([]*fieldCollection, error) {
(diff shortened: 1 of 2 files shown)


entgql/template/mutation_input.tmpl
```diff
@@ -30,27 +30,27 @@ import (
...
+                {{ $structField }} {{ if or (not $n.IsCreate) $e.Optional }}*{{ end }}{{ $e.Type.ID.Type }} `json:{{ $e.Name | camel }}ID`
             {{- else }}
                 {{- if $n.IsCreate }}
                     {{- $structField := print (singular $e.Name | pascal) "IDs" }}
-                    {{ $structField }} []{{ $e.Type.ID.Type }}
+                    {{ $structField }} []{{ $e.Type.ID.Type }} `json:{{ singular $e.Name | camel }}IDs`
                 {{- else }}
-                    {{ $e.MutationAdd }} []{{ $e.Type.ID.Type }}
-                    {{ $e.MutationRemove }} []{{ $e.Type.ID.Type }}
+                    {{ $e.MutationAdd }} []{{ $e.Type.ID.Type }} `json:{{ $e.MutationAdd | decap }}`
```
masseelch, line 52:
```suggestion
                    {{ $e.MutationAdd }} []{{ $e.Type.ID.Type }} `json:{{ $e.MutationAdd | camel }}`
```
Why not this?

masseelch (commented):
This change is backwards incompatible, isn't it? I think we need an option to enable/disable or configure the tag.

entgql/template.go
```diff
@@ -172,6 +174,25 @@ type fieldCollection struct {
 	Mapping []string
 }
 
+// decap takes a string and decapitalizes the first letter
+//
+//	FullName  => fullName
+//	ID  => iD
+func decap(s string) string {
```
a8m, line 181: The generated names are available here: https://github.com/ent/contrib/blob/master/entgql/schema.go#L578-L594. We should bind them to the template data when generating these types (or the <T>WhereInput) either statically or by template function.

philip-peterson:
@masseelch Sure, I can do that. Where would the configuration option live?

entgql/template.go
philip-peterson: Sounds good, I will update.

philip-peterson:
@a8m @masseelch Changes requested have been implemented on this PR, and also I gated the feature behind a feature flag added in this PR: https://github.com/ent/ent/pull/3808 Thanks again for reviewing :)

philip-peterson:
Hi all, @a8m @masseelch is there any way I can help push this PR through?

codelite7:
I've run into this same issue using [gqlgenc](https://github.com/Yamashou/gqlgenc) to generate a client. I can autobind to the input types that entgql creates and the generated client uses them, but since they don't have json tags the request fails when the request body is marshalled out because the field names are not lower camel case. It would be really nice to get this PR merged so that we don't need workarounds to use the generated input types.

elad-aharon:
I'll join the others and mention that PR would be helpful for internal graphql tests when we want to send to a local grapgql client:
```
c.Post(query, &resp, client.Var("input", &input))
```
it's only a matter of `Name` v.s `name` as a struct tag.

As @philip-peterson took care of, PR is backwards compatible.

PR #557, opened 2023-11-16
fix edge comment generation when relayConnection is enabled

Hiroki6:
when the `edgeAnt.RelayConnection` is true, the `Description` field was overwritten as an empty string. Therefore, the comment was not displayed in the  `ein.graphql` schema file.

### DIFF
--- a/entgql/internal/todo/ent.graphql
+++ b/entgql/internal/todo/ent.graphql
@@ -712,7 +712,9 @@ type Todo implements Node {
   init: Map
   custom: [Custom!]
   customp: [Custom]
+  """The parent of the todo"""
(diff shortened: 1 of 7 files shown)


entgql/internal/todo/ent.graphql
```diff
@@ -712,7 +712,9 @@ type Todo implements Node {
   init: Map
   custom: [Custom!]
   customp: [Custom]
+  """The parent of the todo"""
   parent: Todo
+  """The children of the todo"""
```
Hiroki6, line 717: this comment and the comment for groups were not displayed before, even though the parent comment above was displayed.

PR #559, opened 2023-11-23
Multi order by fix

michaelcaulley:
* OrderBy terms for multiOrder + pagination would not properly handle terms that were joined. They would assume the field existed as a column on the table.
* Added a test to verify the change


go.mod
```diff
@@ -41,26 +41,29 @@ require (
...
+	golang.org/x/mod v0.13.0 // indirect
+	golang.org/x/net v0.16.0 // indirect
+	golang.org/x/sys v0.13.0 // indirect
+	golang.org/x/text v0.13.0 // indirect
 	google.golang.org/genproto v0.0.0-20221118155620-16455021b5e6 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace entgo.io/ent => github.com/michaelcaulley/ent v0.11.10-0.20231028174655-603145a879ff
```
michaelcaulley, line 69: This is required for the tests to pass. We'll want to remove it once ent is updated and before merging this pull request.

michaelcaulley: @giautm I updated the pull request and use the master branch for ent.

michaelcaulley:
The tests should pass now if the checks are run again.

entgql/pagination.go
```diff
@@ -218,16 +218,32 @@ func multiPredicate[T any](cursor *Cursor[T], opts *MultiCursorsOptions) (func(*
...
+		getColumnNameForField := func(field string) string  {
+			// The predicate function is executed on query generation time.
+			column := s.C(field)
+			// If there is a non-ambiguous match, we use it. That is because
+			// some order terms may append joined information to query selection.
+			if matches := s.FindSelection(field); len(matches) == 1 {
+				column = matches[0]
+			}
+			return column
+		}
```
michaelcaulley, line 233: using `getColumnNameForField` allows to order by joined data rather than only columns on the table

PR #565, opened 2024-02-15
entgql: ensure order fields are prefixed with type name

a8m:
Also when order field is set on an edge

cc @michaelcaulley
#549


giautm (approved):

michaelcaulley:
Thanks both for getting this merged!

PR #572, opened 2024-04-11
entgql: improve var name collision avoidance in toCursor

henrywoody:
Improves variable name collision avoidance in the `toCursor` method of the pager types in the `gql_pagination` template by suffixing all hard-coded variables names with an underscore.

Address the issue that was the target of #488 and #538. PR #538 fixed the issue for types that start with "O" but broke it for types that start with "Po" (e.g. "Post") or that are two words with the initials "PO" (e.g. "PostOffice"). The issue is also present for types that have the initials "CS" (e.g. "CourseSection"). My fix here puts an underscore as a suffix on all the internal variables so that variable name collision is not possible.

It might be better to opt out of dynamic variables for the argument altogether or updating the variable name generator to accept more context about names to avoid, but this seemed the minimal change to solve the issue. I'll leave it to you guys to decide if a bigger refactor is desired.

### DIFF
--- a/entgql/internal/todo/ent/gql_pagination.go
+++ b/entgql/internal/todo/ent/gql_pagination.go
@@ -476,11 +476,11 @@ func (p *categoryPager) applyFilter(query *CategoryQuery) (*CategoryQuery, error
 }
 
 func (p *categoryPager) toCursor(c *Category) Cursor {
-	cs := make([]any, 0, len(p.order))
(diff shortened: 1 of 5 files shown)


a8m (approved):

a8m:
Thanks for the contribution, @henrywoody 🫶

henrywoody:
Thanks for the quick turnaround @a8m!

PR #577, opened 2024-04-27
feat(entoas): add field descriptions and fix examples

lrstanley:
This PR does two things:
- For all ent fields, the description field is populated in the ogen schema.
  - Had to move away from the `types` map, because this was using the same fields, and changing the contents was impacting all fields of the same type. Don't see any noticeable performance difference -- not sure if this was intentional before or not.
  - Added `ptr[T any](v) *v` to help with code-duplication and should help with all of the pointer-to-base-types issues.
  - Fixes some cases of min/max not being properly applied.
- The `Example()` annotation now works as expected. Previously, adding that annotation to any field, edge, or schema, didn't seem to have any affect.

### DIFF
--- a/entoas/annotation.go
+++ b/entoas/annotation.go
@@ -29,7 +29,7 @@ type (
 		// Groups holds the serialization groups to use on this field / edge.
 		Groups serialization.Groups
 		// OpenAPI Specification example value for a schema field.
-		Example interface{}
+		Example any
 		// OpenAPI Specification schema to use for a schema field.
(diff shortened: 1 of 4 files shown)


lrstanley:
I keep fixing bugs/inconsistencies I run across, and most are captured in this PR. I'd submit multiple PRs for each of these, but many of them are minor, so hopefully the PR isn't too big for review. I will try and avoid adding anything additional if at all possible. If there is anything specific you might want broken out, let me know.

PR #590, opened 2024-06-28
feat: add Extention annotation and add support for x-gen-operation-group ant

prgres:


prgres:
waits for https://github.com/ogen-go/ogen/pull/1269 to be released

PR #591, opened 2024-07-15
protoc-gen-ent: support timestamp and date

simon-wenmouth:
Use of types `google.protobuf.Timestamp` and `google.type.Date` currently panics because they are MessageType but not edges.

> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xae904d]
>
> goroutine 1 [running]:
> main.toEdge(0xc0003d1ba0)
> ... entgo.io/contrib@v0.5.0/entproto/cmd/protoc-gen-ent/main.go:146 +0x8d

The method `isEdge` was updated to recognize these types as fields, and the method `toField` was updated to map them as `field.Time`.

### DIFF
--- a/entproto/cmd/protoc-gen-ent/google/type/date.proto
+++ b/entproto/cmd/protoc-gen-ent/google/type/date.proto
@@ -1,0 +1,52 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


simon-wenmouth:
The tool `protoc-gen-ent` does not currently have support for time fields.  I added mappings for `google.protobuf.timestamp` and `google.type.date` because they have a mapping to `field.Time`.  I did not add a mapping for `google.type.timeofday` (e.g. SQL `time`) because `ent` does not currently support these "civil" date/time types.  If that is of interest, I'm willing to implement this feature and discuss the design with your team.  At a high level I expect this would mean adding `field.CivilDate`, `field.CivilTime` and `field.CivilDateTime` with the underlying structs coming from either `cloud.google.com/go/civil` or `golang-sql/civil`, with corresponding updates to `protoc-gen-ent` afterwards.

I think this PR duplicates in part the work in https://github.com/ent/contrib/pull/447 (March 2023), which has not received feedback.

I welcome your feedback, and look forward to the possibility of the feature getting merged in the future.

Thanks!

PR #592, opened 2024-07-16
protoc-gen-ent: support `entgql.Annotation`

simon-wenmouth:
Added support for `entgql.Annotation` to the `protoc-gen-ent` plugin. It should now be possible to generate an Ent schema and its GraphQL configuration from a Protocol Buffer file.

The `opts.proto` was updated adding `GQL` child MessageTypes to the existing MessageTypes (`Schema`, `Field`, `Edge`).  The associated generated code (`opts.pb.go`) was regenerated using `protoc-gen-go`.

The AST generation code in `schemaast/annotation.go` was updated to know about the `entgql` annotations.  The new method attempts to generate code similar to that in the documentation.

The main package was updated to map the configuration settings in the updated `opts.proto` to the corresponding annotations for the schema, fields, and edges.


simon-wenmouth:
It looks like I should regenerate the `opts.pb.go` using `protoc-gen-go@v1.28.0` and `protoc@v3.19.4`.  I'll update the PR in case this introduces a breaking change for others.

simon-wenmouth:
I've updated the version of `protoc` to match the previous generated version (only a comment changed).

PR #595, opened 2024-07-17
entgql: add annotation Deprecated for deprecated field

luantranminh:


entgql/extension.go
```diff
@@ -244,7 +244,19 @@ func (e *Extension) genSchemaHook() gen.Hook {
...
+				for _, f := range t.DeprecatedFields() {
+					ant, err := annotation(f.Annotations)
+					if err != nil {
+						return err
+					}
+					ant.Directives = append(ant.Directives, Deprecated(f.DeprecationReason()))
+					if f.Annotations == nil {
+						f.Annotations = make(map[string]interface{})
+					}
+					f.Annotations["EntGQL"] = ant
```
a8m, line 257: Make sure the field does not already have such annotation to avoid duplicate directives.

a8m, line 257:
```suggestion
					f.Annotations[ant.Name()] = ant
```

entgql/internal/todo/gen_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"path/filepath"
+	"testing"
+
+	"entgo.io/contrib/entgql"
+	"entgo.io/ent/entc"
+	"entgo.io/ent/entc/gen"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGeneratedSchema(t *testing.T) {
```
a8m, line 28: Nice, but don't we already test this here: https://github.com/ent/contrib/blob/master/entgql/schema_test.go ?
I don't mind keeping it here.

cc @giautm

luantranminh: schema_test runs tests by calling `buildTypes`. I added logic to `schemaHook`, which executes before `buildTypes`

giautm (approved):

giautm:
Good Job, welcome a new contributor.

PR #597, opened 2024-07-20
entgql: make connection node non-null

yukukotani:
https://github.com/ent/ent/issues/3998

## What's changed

Generates `node` field in connection edge types as non-null types.

## Background

According to [the spec](https://relay.dev/graphql/connections.htm), `node` field can be non-null.

> An “Edge Type” must contain a field called node. This field must return either a Scalar, Enum, Object, Interface, Union, **or a Non-Null wrapper around one of those types**. Notably, this field cannot return a list.
> https://relay.dev/graphql/connections.htm#sel-FAHFFDCAACEP0xF

On the other hand, `edges` field in connection types are defined as just *a list type that wraps an edge type*, which means `[Edge]` but not `[Edge]!`.

> A “Connection Type” must contain a field called edges. This field must return **a list type that wraps an edge type**, where the requirements of an edge type are defined in the “Edge Types” section below.
> https://relay.dev/graphql/connections.htm#sel-FAFFFDCAACCzEsY

So this PR only changes `node` field and keeps `edges` field as it is.

## Concerns

My concern is whether we should consider this a breaking change.

GraphQL API clients can accept the transition from nullable to non-null fields without any change or breaking.

How about servers? I'm not sure this change would break server implementations using ent.

If so, we can introduce this change as an option.

### DIFF
--- a/entgql/internal/todo/ent.graphql
+++ b/entgql/internal/todo/ent.graphql
@@ -165,7 +165,7 @@ type CategoryEdge {
   """
   The item at the end of the edge.
   """
-  node: Category
+  node: Category!
   """
   A cursor for use in pagination.
   """
@@ -373,7 +373,7 @@ type FriendshipEdge {
   """
   The item at the end of the edge.
   """
-  node: Friendship
+  node: Friendship!
   """
   A cursor for use in pagination.
   """
@@ -469,7 +469,7 @@ type GroupEdge {
   """
   The item at the end of the edge.
   """
-  node: Group
+  node: Group!
   """
   A cursor for use in pagination.
   """
@@ -561,7 +561,7 @@ type OneToManyEdge {
   """
   The item at the end of the edge.
   """
-  node: OneToMany
+  node: OneToMany!
   """
   A cursor for use in pagination.
   """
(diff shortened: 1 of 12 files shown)


freb:
Is there anything holding this up? This issue really creates a lot of unnecessary work for TypeScript users.

yukukotani:
idk, I'm ready for getting review!

a8m:
> Is there anything holding this up? This issue really creates a lot of unnecessary work for TypeScript users.

I believe this behavior should be config-based. Also, this can already be achieved today using a [schema hook](https://github.com/ent/contrib/blob/master/entgql/extension.go#L71-L78), so maybe we should provides a builtin hook for this. 

@giautm, your thoughts?

giautm:
> @giautm, your thoughts?

Nullable to Non-nullable is a safe change, and should not consider as breaking changes. It just a bit annoying when update the generated code. The important thing we need check that is no cases entgql will return null for some nodes.

The original implement was from Facebook respect, but when the spec go out community and grow, it doesn't correct anymore. So, I prefer to go with new Relay spec.

PR #599, opened 2024-07-31
entgql: fix template rendering for pulid example

kdevo:
Fixes a regression introduced during an old commit by re-enabling template rendering of ent/schema/pulid/template/pulid.tmpl


a8m (approved):
💯 

Thanks for the fix.

PR #602, opened 2024-10-10
entgql: add nodes to connection

eiixy:


nicolerenee:
@a8m (sorry for the direct tag) any chance of getting a review or feedback on this, we would also like to be able to use this feature.

PR #603, opened 2024-10-12
Use dynamic Value/GetValue method name in pagination with edge order fields

henrywoody:
Addresses https://github.com/ent/ent/issues/4227


a8m (approved):
💯 

Thanks @henrywoody

henrywoody:
@a8m For sure, give me one moment to fix the tests

henrywoody:
Okay @a8m got it settled now, thanks for the quick review!

PR #608, opened 2024-12-22
ability to add additonal fields and funcs to mutation input

khanakia:
Refer: https://github.com/ent/ent/pull/4281

I am building a Media Extension need to set Fields on e.g PostMutation via `saas/gen/ent/gql_mutation_input.go`

So instead of copying the whole template file and overriding, we should be able to override within the default template from the vendor directory.

and also this way multiple custom extensions can add additional fields and make it modular.

### DIFF
--- a/entgql/template/mutation_input.tmpl
+++ b/entgql/template/mutation_input.tmpl
@@ -55,6 +55,12 @@ import (
                 {{- end }}
             {{- end }}
         {{- end }}
+
+        {{- with $tmpls := matchTemplate "helper/gql_mutation_input/model/fields/*"  }}
+            {{- range $tmpl := $tmpls }}
+                {{- xtemplate $tmpl $n }}
+            {{- end }}
+        {{- end }}
     }
 
     // Mutate applies the {{ $input }} on the {{ $n.MutationName }} builder.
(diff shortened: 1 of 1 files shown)


entgql/template/mutation_input.tmpl
```diff
@@ -55,6 +55,12 @@ import (
                 {{- end }}
             {{- end }}
         {{- end }}
+
+        {{- with $tmpls := matchTemplate "helper/gql_mutation_input/model/fields/*"  }}
```
a8m, line 59:
```suggestion
        {{- with $tmpls := matchTemplate "helper/gql_mutation_input/create/fields/*"  }}
```

a8m, line 59: You have two types of input objects here. Create or Update. 

If already support this, let's add both?

a8m: You can use `$n.IsCreate` to check if it's `create/fields` or `update/fields`.

khanakia: Yes let's the user developer decide that themselves so it will give them freedom sometimes I could have need the same field to be available in both Create and Update input so I will have to duplicate the code this way if we separate them


User can decided themselves like this
```go
{{ define "helper/gql_mutation_input/model/fields/saasmedia_mutation" }}
	{{- $n := $ }}
		{{- if $n.Annotations.SaasMedia }}
			{{- range $e := $n.Annotations.SaasMedia.Fields }}
				{{- if $n.IsCreate }}

					{{ print ($e.StructField | singular) }}IDs []{{ $e.Type }}
					
					{{- else }}

						{{- $MutationAdd := print "Add" $structField }}
						{{ $MutationAdd }} []string
				{{- end }}
				
			{{- end }}
		{{- end }}
{{ end }}
```

a8m, line 59:
```suggestion
        {{- with $tmpls := matchTemplate "helper/gql_mutation_input/fields/*"  }}
```

a8m (approved):
See minor comment

PR #607, opened 2024-12-18
chore: bump go to 1.23.4 and gqlgen to v0.17.51

michaelcaulley:
# Summary
* bump go to 1.23.4 and gqlgen to v0.17.51
   * it appears that gqlgen v0.17.52 breaks field generation with ent. See: https://github.com/99designs/gqlgen/commit/f81239e47a7ebaf1ab276689344e1189f83ac0a0

PR #609, opened 2024-12-30
Update 99designs/gqlgen@v0.17.62 and vektah/gqlparser/v2@v2.5.21

StevenACoffman:
### Update gqlgen and gqlparser

This updates gqlgen and gqlparser to the latest versions [github.com/99designs/gqlgen@v0.17.61](https://github.com/99designs/gqlgen/releases/tag/v0.17.61) and [github.com/vektah/gqlparser/v2@v2.5.21](https://github.com/vektah/gqlparser/releases/tag/v2.5.21)

---

In https://github.com/ent/contrib/pull/607/files github.com/99designs/gqlgen was moved from v0.17.48 to v0.17.51.

I see that in that PR it has a comment that says that gqlgen version v0.17.52 has an issue with ent field collection. It blames commit: https://github.com/99designs/gqlgen/commit/f81239e47a7ebaf1ab276689344e1189f83ac0a0

The blamed gqlgen commit was merged https://github.com/99designs/gqlgen/pull/3287
This was to fix https://github.com/99designs/gqlgen/issues/3285 by reverting https://github.com/99designs/gqlgen/pull/3203

Currently gqlgen v0.17.61 is the latest. It is not clear to me if there remains any problem with gqlgen and ent field collection.
Please point me to an open issue in gqlgen (or PR) describing (or fixing) the ent field collection problem.


StevenACoffman:
Hmmm, I'm not sure if my local environment is entirely what was expected. I appear to be finding the required local setup by process of elimination. So far:

```
export GOBIN=~/go/bin
go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28.0
go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2.0
cd entproto/cmd/protoc-gen-entgrpc
go install
cd ../../..
cd entproto/cmd/protoc-gen-ent
go install
cd ../../..
go generate ./...
```

StevenACoffman:
Ok, I am getting go generate failures when I run [nektos/act](https://github.com/nektos/act) to simulate the remote GitHub action failures:
```
act -j generate
```
This provides this minor output tweak:
```
| diff --git a/entgql/internal/todogotype/generated.go b/entgql/internal/todogotype/generated.go
| index dd27c93..d9738af 100644
| --- a/entgql/internal/todogotype/generated.go
| +++ b/entgql/internal/todogotype/generated.go
| @@ -20630,7 +20630,7 @@ func (ec *executionContext) marshalOInt2ᚖint(ctx context.Context, sel ast.Sele
|  	return res
|  }
|
| -func (ec *executionContext) unmarshalOMap2map(ctx context.Context, v any) (map[string]any, error) {
| +func (ec *executionContext) unmarshalOMap2map(ctx context.Context, v any) (map[string]interface{}, error) {
|  	if v == nil {
|  		return nil, nil
|  	}
| @@ -20638,7 +20638,7 @@ func (ec *executionContext) unmarshalOMap2map(ctx context.Context, v any) (map[s
|  	return res, graphql.ErrorOnPath(ctx, err)
|  }
|
| -func (ec *executionContext) marshalOMap2map(ctx context.Context, sel ast.SelectionSet, v map[string]any) graphql.Marshaler {
| +func (ec *executionContext) marshalOMap2map(ctx context.Context, sel ast.SelectionSet, v map[string]interface{}) graphql.Marshaler {
|  	if v == nil {
|  		return graphql.Null
|  	}
[Continuous Integration/generate]   ❌  Failure - Main Check generated files
[Continuous Integration/generate] exitcode '1': failure
[Continuous Integration/generate]   🐳  docker exec cmd=[/opt/acttoolcache/node/18.20.5/arm64/bin/node /var/run/act/workflow/hashfiles/index.js] user= workdir=
[Continuous Integration/generate] 🏁  Job failed

```

StevenACoffman:
Ok, moved to gqlgen v0.17.62 to make any/interface consistent

StevenACoffman:
Running `go test ./...` causes the `go.mod` to add `github.com/google/addlicense`, but `go mod tidy` (which is in the `go generate ./...` ci job) removes it from a direct dependency (and moves it to an indirect). This causes a race condition if you locally run the ci github actions in the wrong order using [nektos/act](https://github.com/nektos/act). Previously, I committed the `go.mod` and was puzzled over why the CI checks kept flagging changing `go.mod` / `go.sum`. To work around that, I'm adding a `tools.go` file (which will be useless in Go 1.24).

The other discrepancy is that my locally running `go generate ./...` seems to do something different for me using Go 1.23.4 on darwin/arm64 vs. when I run `act -j generate --container-architecture linux/amd64`. The remaining difference is between `any` and `interface{}` which I find surprising.

```
diff --git a/entgql/internal/todogotype/generated.go b/entgql/internal/todogotype/generated.go
index 76712c9..0e24fdf 100644
--- a/entgql/internal/todogotype/generated.go
+++ b/entgql/internal/todogotype/generated.go
@@ -20258,7 +20258,7 @@ func (ec *executionContext) marshalOInt2ᚖint(ctx context.Context, sel ast.Sele
 	return res
 }
 
-func (ec *executionContext) unmarshalOMap2map(ctx context.Context, v any) (map[string]any, error) {
+func (ec *executionContext) unmarshalOMap2map(ctx context.Context, v any) (map[string]interface{}, error) {
 	if v == nil {
 		return nil, nil
 	}
@@ -[20](https://github.com/ent/contrib/actions/runs/12564902369/job/35044421968?pr=609#step:11:21)266,7 +20266,7 @@ func (ec *executionContext) unmarshalOMap2map(ctx context.Context, v any) (map[s
 	return res, graphql.ErrorOnPath(ctx, err)
 }
 
-func (ec *executionContext) marshalOMap2map(ctx context.Context, sel ast.SelectionSet, v map[string]any) graphql.Marshaler {
+func (ec *executionContext) marshalOMap2map(ctx context.Context, sel ast.SelectionSet, v map[string]interface{}) graphql.Marshaler {
 	if v == nil {
 		return graphql.Null
 	}
```

StevenACoffman:
So there's a slight discrepancy in the `go.sum` between my local `go mod tidy` and the `nektos/act -j generate` one, so I ran `go clean --modcache && go mod tidy` and that seemed to clear it up.

a8m (approved):

a8m:
Thanks for your contribution, @StevenACoffman 🙏🏻

michaelcaulley:
@StevenACoffman, adding an additional thanks for helping with this. I really appreciate all your work on gqlgen!

entgql/internal/todo/gqlgen.yml
```diff
@@ -9,6 +9,7 @@ schema:
 
 resolver:
   layout: follow-schema
+  preserve_resolver: true
```
michaelcaulley, line 12: Did we want this to be true or false?

StevenACoffman: I picked true to be safe, because it depends. If there is customized code in those resolvers, then we want to preserve them (true), so we do not want to overwrite them. If there is nothing customized, then I guess it's harmless to preserve them, but we could set it to false.

I took this to be more of an example, as you people aren't actively changing these (or the schema) on a regular basis.

StevenACoffman:
@michaelcaulley Hey, so we should probably bump gqlgen again after this, or else swap the models to use `any` instead of `interface{}` or there is a potential for inconsistent results prior to v0.17.63

michaelcaulley:
If it's set to true, new ent schemas will not be added when running `go generate .`

If helpful, I can create a PR to update preserve_resolver to false and bump gqlgen.

@a8m or @giautm , can you confirm my thinking for `preserve_resolver:` ?

PR #611, opened 2025-01-15
entgql: support global id feature from ent

masseelch:


entgql/internal/todoglobalid/ent/gql_node.go
```diff
@@ -0,0 +1,472 @@
...
+
+type nodeOptions struct {
+	nodeType func(context.Context, int) (string, error)
+}
+
+// Each node has a range of 1<<32 ids. Pre-computing a map of node starting
+// value to node type ensures node type resolution happens in O(1) complexity.
+var nodeTypes = func() map[int]string {
+	var is gen.IncrementStarts
+	if err := json.Unmarshal([]byte(internal.IncrementStarts), &is); err != nil {
```
a8m, line 123: If you don't validate it, why not using `map[string]int64` instead?

a8m (approved):

PR #615, opened 2025-02-19
fix: Allow empty where input

YiminArava4508:
Deleted the line of code in gql_where_input.go that would error you out when a predicate is empty,


YiminArava4508:
Sorry, I chose the wrong source to open a pr on.

PR #619, opened 2025-05-01
Entgql: Determine required fields by Nillable not Optional

freb:
Optional fields are fields not required for entity creation. However, unless those fields are also Nillable, querying the entity will always provide a value for optional fields. Nillable, not Optional, should determine which type fields are marked as required in the generated GraphQL schema.


freb:
I misunderstood how Optional() works. I didn't realize these fields were created as nullable in the database and that the gqlgen code receives the potentially null value and not the zero value as accessible from the generated ent type.

PR #625, opened 2026-01-29
perf(entgql): avoid COUNT(*) for pageInfo-only connection queries

tutranngoc:
PR Description: Avoid unnecessary COUNT(*) query when resolving pageInfo

When querying a connection with `pageInfo`, we always execute an extra SQL query (`SELECT COUNT(*)`). This is not required for computing pageInfo and adds avoidable dat

The `build()` call in:
```go
conn.build(nodes, pager, after, first, before, last)
```
already computes cursor fields (`startCursor`, `endCursor`) and correctly determines pagination flags (`hasNextPage`, `hasPreviousPage`) based on the paging arguments and returned nodes.

Currently we also derive flags using `TotalCount`, which forces a `COUNT(*)`:
```go
conn.PageInfo.HasNextPage = first != nil && conn.TotalCount > 0
conn.PageInfo.HasPreviousPage = last != nil && conn.TotalCount > 0
```

This makes every request that includes `pageInfo` pay for a `COUNT(*)`, even when callers don’t request `totalCount` and we don’t actually need it.

What this PR changes:
- Stops triggering SELECT COUNT(*) when it’s not needed for pageInfo.
- Relies on the existing build() logic for pageInfo instead of TotalCount.

Why it matters
- Reduces database load by removing redundant COUNT(*) queries.
- Helps prevent query amplification in nested resolvers (a common source of N+1-like behavior).
- Improves latency for list queries that request pageInfo but not totalCount.

Example query affected:
```graphql
variants(first: $firstVariant, after: $afterVariant) {
    edges {
      cursor
      node {
        id
        title
        barcode
        baseID
        costPrice
        height
      }
    }
    pageInfo {
      hasNextPage
      endCursor
    }
  }
```
image

### DIFF
--- a/entgql/internal/todo/ent/gql_pagination.go
+++ b/entgql/internal/todo/ent/gql_pagination.go
@@ -292,7 +292,7 @@ func (bp *BillProductQuery) Paginate(
 	}
 	conn := &BillProductConnection{Edges: []*BillProductEdge{}}
 	ignoredEdges := !hasCollectedField(ctx, edgesField)
-	if hasCollectedField(ctx, totalCountField) || hasCollectedField(ctx, pageInfoField) {
+	if hasCollectedField(ctx, totalCountField) {
 		hasPagination := after != nil || first != nil || before != nil || last != nil
 		if hasPagination || ignoredEdges {
 			c := bp.Clone()
@@ -590,7 +590,7 @@ func (c *CategoryQuery) Paginate(
 	}
 	conn := &CategoryConnection{Edges: []*CategoryEdge{}}
 	ignoredEdges := !hasCollectedField(ctx, edgesField)
-	if hasCollectedField(ctx, totalCountField) || hasCollectedField(ctx, pageInfoField) {
+	if hasCollectedField(ctx, totalCountField) {
 		hasPagination := after != nil || first != nil || before != nil || last != nil
 		if hasPagination || ignoredEdges {
 			c := c.Clone()
@@ -981,7 +981,7 @@ func (f *FriendshipQuery) Paginate(
 	}
 	conn := &FriendshipConnection{Edges: []*FriendshipEdge{}}
 	ignoredEdges := !hasCollectedField(ctx, edgesField)
-	if hasCollectedField(ctx, totalCountField) || hasCollectedField(ctx, pageInfoField) {
+	if hasCollectedField(ctx, totalCountField) {
 		hasPagination := after != nil || first != nil || before != nil || last != nil
 		if hasPagination || ignoredEdges {
 			c := f.Clone()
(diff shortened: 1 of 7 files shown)


tutranngoc:
@a8m @giautm Please check this pull request for me 🙇‍♂️🙇‍♂️🙇‍♂️

giautm:
@tutranngoc fyi, you can use entcache package to cache query result of count(). So it wont hit database twice.

https://github.com/ariga/entcache

tutranngoc:
> @tutranngoc fyi, you can use entcache package to cache query result of count(). So it wont hit database twice.
> 
> https://github.com/ariga/entcache

@giautm Thank you for sharing, however, my preference is to eliminate the need for a `select count` when using `pageInfo`, even during the first query. Querying for `count` when using `pageInfo` is unnecessary.

isakstenstrom:
I don't think this PR handles all edge cases for the pagination. If you create a query that looks like this:

```graphql
variants(first: $firstVariant, after: $afterVariant) {
    pageInfo {
      hasNextPage
      endCursor
    }
  }
```

then we will not populate the pageInfo object correctly. If we request pagination info without any edges, we still need to do the count.

I solved the same problem using an external template, where I used this logic instead:

```go
ignoredEdges := !hasCollectedField(ctx, edgesField)
needTotalCount := hasCollectedField(ctx, totalCountField)
needPageInfo := hasCollectedField(ctx, pageInfoField)
hasPagination := after != nil || first != nil || before != nil || last != nil
if (needTotalCount && hasPagination) || (ignoredEdges && (needTotalCount || needPageInfo)) {
	c := {{ $r }}.Clone()
	{{- /* Clear the selection fields before counting to avoid generating invalid queries. */}}
	c.ctx.Fields = nil
	if conn.TotalCount, err = c.Count(ctx); err != nil {
		return nil, err
	}
	conn.PageInfo.HasNextPage = first != nil && conn.TotalCount > 0
	conn.PageInfo.HasPreviousPage = last != nil && conn.TotalCount > 0
}
```

tutranngoc:
> I don't think this PR handles all edge cases for the pagination. If you create a query that looks like this:
> 
> ```graphql
> variants(first: $firstVariant, after: $afterVariant) {
>     pageInfo {
>       hasNextPage
>       endCursor
>     }
>   }
> ```
> 
> then we will not populate the pageInfo object correctly. If we request pagination info without any edges, we still need to do the count.
> 
> I solved the same problem using an external template, where I used this logic instead:
> 
> ```go
> ignoredEdges := !hasCollectedField(ctx, edgesField)
> needTotalCount := hasCollectedField(ctx, totalCountField)
> needPageInfo := hasCollectedField(ctx, pageInfoField)
> hasPagination := after != nil || first != nil || before != nil || last != nil
> if (needTotalCount && hasPagination) || (ignoredEdges && (needTotalCount || needPageInfo)) {
> 	c := {{ $r }}.Clone()
> 	{{- /* Clear the selection fields before counting to avoid generating invalid queries. */}}
> 	c.ctx.Fields = nil
> 	if conn.TotalCount, err = c.Count(ctx); err != nil {
> 		return nil, err
> 	}
> 	conn.PageInfo.HasNextPage = first != nil && conn.TotalCount > 0
> 	conn.PageInfo.HasPreviousPage = last != nil && conn.TotalCount > 0
> }
> ```

Thanks for your post, I’ll give it a try.

PR #628, opened 2026-02-25
fix(entgql): method wrappers and enum prefixes for split-package compatibility

codelite7:
## Summary

This branch adds compatibility fixes for entgql when using split-package code generation (per-entity packages).

- **Fix enum variable prefixes**: prefix enum map variables with entity name to avoid redeclarations across packages
- **Restore sub-package compatibility**: restore enum methods to meta template for cross-package use
- **Add method wrappers to edge_entity template**: relay connection edges need wrappers when types are in separate packages
- **Add SetInput method wrappers to mutation_input_entity template**: mutation input types need wrappers for cross-package access
- **Restore unexported config field**: fix mutation input entity template regression

These fixes are needed when using split/sharded code generation where each entity lives in its own package, causing symbol redeclaration and cross-package method resolution issues.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


codelite7:
Opened against wrong repo, will be handled in the fork.

PR #631, opened 2026-05-15
feat(entgql): expression-aware ORDER BY + cursor predicates

codelite7:
## Summary

Adds an optional `OrderFieldExpr` annotation that swaps a SQL expression for the raw column in ORDER BY emission and cursor comparisons. Motivated by composite indexes of the form `(left("col", 256), id)` — needed for cursor-based pagination on unbounded text columns that would otherwise overflow the 2704-byte btree max tuple size. The left-expression index only helps if the query also uses the same expression; this PR teaches entgql how to.

## New surface

- `entgql.OrderFieldExpr("left(\"name\", 256)")` — schema annotation paired with `entgql.OrderField` on the same field.
- `entgql.CursorsPredicateExpr` — variant of `CursorsPredicate` that compares against a raw SQL expression instead of a quoted column.
- Generated `<Entity>OrderField` struct gains an `expression` string field. When set, ORDER BY emission writes the expression raw and `toTerm` calls `sql.OrderExprFunc` instead of ent's `By<Field>` helper. `column` is kept for SELECT list inclusion so cursor values can still be read back per row.

Templates updated for both monolithic (`pagination.tmpl`) and entity-split (`pagination_entity.tmpl`) generation modes. **Behavior is unchanged for any field that doesn't carry `OrderFieldExpr`** — the existing column path is preserved.

## Why this is needed

Postgres btree indexes have a 2704-byte max tuple size. A composite `(description, id)` index on an unbounded `text` column fails to build the moment a row's `description` exceeds that. A `(left("description", 256), id)` index sidesteps the limit, but `ORDER BY description, id` won't use it — the planner needs an exact expression match. Without this PR, users either skip indexing those columns (slow sort) or build indexes that no query can use.

## Test plan

- [x] All existing entgql tests pass (`go test ./...`)
- [x] New unit tests for `CursorsPredicateExpr` covering ASC/DESC, nil-value cursors (ID-only fallback), and nil-cursor inputs
- [x] New unit tests for `OrderFieldExpr` annotation and Merge behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)


codelite7:
Opened against wrong base; reopening against codelite7/contrib fork.

PR #636, opened 2026-06-16
entgql: schema-qualify M2M join table in totalCount loader

michaelcaulley:
The gql_collection.go totalCount loader for M2M edges builds the join table via `sql.Table(...)` but never calls `joinT.Schema(...)`. With the `sql/schemaconfig` feature enabled, this emits an unqualified join table in the generated SQL while the rest of the query is correctly schema- qualified, producing `relation does not exist` errors at runtime.

Hook the same `dialect/sql/query/eagerloading/join/*` template extension point that ent core's query.tmpl uses for the loader path, so the join table picks up its schema from `schemaConfig`. No-op when the feature is disabled (verified: all 6 integration tests regenerate byte-identically).


a8m:
cc @chinhnguyenh for review too.

chinhnguyenh:
Look good to me.
It would be nice if you can add a small example project for it.

michaelcaulley:
> Look good to me. It would be nice if you can add a small example project for it.

Hi, @chinhnguyenh, are you on Discord? If so, can I reach out there? I have a few questions about the best way to add a test. If we want to create a new multischema project, we'll need MySQL to test it. Otherwise, we can limit the test to reading the generated code. 

Edit: I've added each option as a separate commit. If you'd like to discuss either option, please feel free to message me on Discord.

chinhnguyenh:
> > Look good to me. It would be nice if you can add a small example project for it.
> 
> Hi, @chinhnguyenh, are you on Discord? If so, can I reach out there? I have a few questions about the best way to add a test. If we want to create a new multischema project, we'll need MySQL to test it. Otherwise, we can limit the test to reading the generated code.
> 
> Edit: I've added each option as a separate commit. If you'd like to discuss either option, please feel free to message me on Discord.

We don't need to test it with MySQL, verifing the generated code is enough.
Please rename the project to `todoschema` to keep consistency with others.

entgql/template_test.go
```diff
@@ -294,3 +298,52 @@ func TestFilterFields(t *testing.T) {
...
+
+	out, err := os.ReadFile(filepath.Join(target, "gql_collection.go"))
+	require.NoError(t, err)
+	return string(out)
+}
+
+// TestCollection_M2MTotalCountSchemaConfig verifies that the M2M totalCount
+// loader schema-qualifies its join table when the sql/schemaconfig feature is
+// enabled, and leaves it unqualified otherwise.
+func TestCollection_M2MTotalCountSchemaConfig(t *testing.T) {
```
chinhnguyenh, line 327: Create `gen_test.go` in the example project and create tests there instead.

michaelcaulley: If we're testing only the generated code, do we need the new project? Can I drop my last commit?

chinhnguyenh: The example project is what CI regenerates and diffs. The test only reads the output, but the project is what keeps that output committed, compilable, and in sync with the templates.

chinhnguyenh: Anyway, it's just a nice-to-have.

a8m (approved):

